PR #81, opened 2018-04-24
Enable lint and tests for notebooks in example folder

leezu:
## Description ##
This PR enables lint checks and basic tests for the notebooks in the example folder. In particular it is tested that every notebook runs without errors. For that the PR enables the pytest extension [nbsmoke](https://github.com/pyviz/nbsmoke).

Alternatively we could also test that the output of each cell exactly matches the stored output by using https://github.com/computationalmodelling/nbval . This may be problematic as some of our outputs could be non-deterministic.


cclauss:
Please resolve conflicts

szha:
the goal of this PR is already covered in the current CI.

PR #92, opened 2018-04-28
Fix sphinx docs

leezu:
## Description ##
This PR fixes all documentation that causes warnings when building the documentation. It also switches the CI to only pass if the documentation doesn't cause warnings.

The rationale is to make sure that we never create any errors in our documentation by accident (broken links, missing files, etc.).


mli:
Job PR-92/5 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-92/5/index.html

PR #95, opened 2018-05-02
[Proposal] Move/rename files with history

cgraywang:
## Description ##
This is a proposal. I suggest, In the future, if we decide to move or rename files, we can consider to preserve the history of the file. In this case, it is easier for us to keep track of the history versions of the file and commit history, in case we need to revert or other operations needed.

In order to do this, please follow the following example.

If you want the project's history to look as though all files have always been in the directory foo/bar, then you need to do a little surgery. Use git filter-branch with the "tree filter" to rewrite the commits so that anywhere foo/bar doesn't exist, it is created and all files are moved to it:

```
git filter-branch --prune-empty --tree-filter '
if [ ! -e foo/bar ]; then
    mkdir -p foo/bar
    git ls-tree --name-only $GIT_COMMIT | xargs -I files mv files foo/bar
fi'
```
Now the history will be recorded as if all files were always located in foo/bar.

### Essentials ###
- [ ] Code is well-documented

### Changes ###

## Comments ##

### DIFF
--- a/scripts/language_model/word_language_model.py
+++ b/scripts/language_model/word_language_model.py
@@ -151,7 +151,6 @@ print(args)
 # Build the model
 ###############################################################################
 
-
 ntokens = len(vocab)
 
 if args.weight_dropout > 0:


mli:
Job PR-95/1 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-95/1/index.html

mli:
Job PR-95/2 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-95/2/index.html

leezu:
Could you give some example where git did not track the history of file renames in our project? I was under the impression that the information should be tracked already.

Are you referring to the rewriting the history of the master branch, or of some commits that you plan to PR?

I [quote](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) the following about filter-branch (highlighting mine):

> There is another history-rewriting option that you can use if you need to rewrite a larger number of commits in some scriptable way – for instance, changing your email address globally or removing a file from every commit. **The command is filter-branch, and it can rewrite huge swaths of your history, so you probably shouldn’t use it unless your project isn’t yet public and other people haven’t based work off the commits you’re about to rewrite**. However, it can be very useful. You’ll learn a few of the common uses so you can get an idea of some of the things it’s capable of.

cgraywang:
when moving files, the history is lost. For example:

https://github.com/dmlc/gluon-nlp/blob/master/docs/examples/language_model/language_model.ipynb

https://github.com/dmlc/gluon-nlp/blob/master/contributor.rst

leezu:
It turns out we just need to run:

`git log --follow PATH` to follow renames through history. Unfortunately [github doesn't do that by default.](https://stackoverflow.com/questions/5646174/how-to-make-github-follow-directory-history-after-renames/) To always enable the `--follow` option one may set `git config log.follow true`.

PR #118, opened 2018-05-23
Fix unknown token handling for vocab

leezu:


mli:
Job PR-118/3 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-118/3/index.html

gluonnlp/vocab.py
```diff
@@ -304,6 +304,11 @@ def set_embedding(self, *embeddings):
                 'The argument `embeddings` must be an instance or a list of instances of ' \
                 '`gluonnlp.embedding.TokenEmbedding`.'
 
+        assert all([embs.unknown_token for embs in embeddings]) or \
+            all([not embs.unknown_token for embs in embeddings]), \
```
szha: nit: a list is not needed here. a generator suffices.

leezu: Lists will generate nicer error messages during tests if the assertion fails. But lets change it to a generator as it will improve performance.

mli:
Job PR-118/4 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-118/4/index.html

gluonnlp/vocab.py
szha: on second thought, it's probably the same as it's hard to imagine user feeding 1mm different embeddings here...

PR #128, opened 2018-06-01
Don't treat language model datasets special

leezu:
Currently WikiText datasets inherit from LanguageModelDataset which provides them with some language model specific transformers for generating batches. However, WikiText can also be used for other purposes. This PR removes the LanguageModelDataset class and moves the language model specific batch generation code to data.batchify.


szha:
we should update the version to 0.3 because of the api change.

mli:
Job PR-128/5 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-128/5/index.html

szha:
@cgraywang there's no change for Wikitext datasets in terms of module. File name doesn't affect where the class resides because of the import in `__init__.py`. Do you have other suggestions on the code?

szha:
cc'd @eric-haibin-lin who might be affected.

cgraywang:
I know there is no affect on using the dataset. I am not sure the necessity
of changing the file' name. For example, sentiment analysis is also a
separate file.

cgraywang:
If that is the case, sentiment analysis should be also renamed: https://github.com/dmlc/gluon-nlp/pull/130 .

PR #129, opened 2018-06-02
Add convolutional encoder and highway layer

cgraywang:
## Description ##
1. Add convolutional encoder
2. Add highway layer and higway bias initializer
3. Corresponding docs

### Essentials ###
- [ ] Code is well-documented

### Changes ###

## Comments ##


mli:
Job PR-129/2 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-129/2/index.html

mli:
Job PR-129/13 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-129/13/index.html

gluonnlp/model/char_encoder.py
```diff
@@ -0,0 +1,144 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+"""Convolutional character encoder."""
+
+from __future__ import absolute_import
```
eric-haibin-lin: stupid question: why do we need absolute_import?

gluonnlp/model/char_encoder.py
```diff
@@ -0,0 +1,144 @@
...
+    ngram_filter_sizes: Tuple[int]
+        The size of each convolutional layer,
+        and len(ngram_filter_sizes) equals to the number of convolutional layers.
+    conv_layer_activation: nn.Activation
+        Activation function to be used after convolutional layer.
+    num_highway: int
+        The number of layers of the Highway layer.
+    output_size: int
+        The output dimension after conducting the convolutions and max pooling,
+        and apply highways, as well as linear projection.
```
eric-haibin-lin: grammar: ... after conducting convolutions and max pooling, and applying highways ...

gluonnlp/model/char_encoder.py
```diff
@@ -0,0 +1,144 @@
...
+    def forward(self, inputs, mask=None): # pylint: disable=arguments-differ
+        r"""
+        Forward computation for char_encoder
+
+        Parameters
+        ----------
+        inputs: NDArray
+            The input tensor is of shape `(seq_len, batch_size, embedding_size)` TNC.
+        mask: NDArray
+            The mask applied to the input of shape `(seq_len, embedding_size)`
```
eric-haibin-lin: Why is mask 2-D?

gluonnlp/model/char_encoder.py
cgraywang: from __future__ import absolute_import means that if you import gluonnlp, Python will always look for a top-level gluonnlp module, rather than current_package.gluonnlp

gluonnlp/model/char_encoder.py
cgraywang: Thanks

gluonnlp/model/char_encoder.py
cgraywang: because we are using ``inputs * mask.expand_dims(-1)`` , the mask will be broadcasted along the embedding dimension.

cgraywang: Yes it is a mistake, thanks. Updated

mli:
Job PR-129/15 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-129/15/index.html

docs/api/model.rst
```diff
@@ -21,6 +21,22 @@ Language Modeling
     StandardRNN
     get_model
 
+Character Encoder
+-----------------
+
+.. autosummary::
+    :nosignatures:
+
+    CharacterEncoder
```
szha: Would it be appropriate to name it as ConvolutionEncoder? We don't have any module that dictates the form of the input data so far.

szha: this encoder might be useful to encode sequences of other granularity

gluonnlp/model/highway.py
```diff
@@ -0,0 +1,108 @@
...
+            self.hnet = nn.HybridSequential()
+            with self.hnet.name_scope():
+                for _ in range(self._num_layers):
+                    self.hnet.add(nn.Dense(units=self._input_size * 2, in_units=self._input_size,
+                                           use_bias=True, flatten=False))
+            self._activation = activation
+
+    def set_bias(self):
+        for _, layer in enumerate(self.hnet):
+            layer.bias.data()[self._input_size:] = 1
```
szha: what does this do and why is it a public method?

szha: consider using a bias_initializer in constructor for the same effect.

gluonnlp/model/highway.py
```diff
@@ -0,0 +1,108 @@
...
+        current_input = inputs
+        for _, layer in enumerate(self.hnet):
+            projected_input = layer(current_input)
+            linear_transform = current_input
+            nonlinear_transform, transform_gate = projected_input.split(num_outputs=2, axis=-1)
+            nonlinear_transform = self._activation(nonlinear_transform)
+            transform_gate = transform_gate.sigmoid()
+            current_input = (1 - transform_gate) * linear_transform + \
+                            transform_gate * nonlinear_transform
+        return current_input
```
szha: seems ok to be a hybrid block

gluonnlp/model/char_encoder.py
```diff
@@ -0,0 +1,145 @@
...
+        self._embed_size = embed_size
+        self._num_filters = num_filters
+        self._ngram_filter_sizes = ngram_filter_sizes
+        self._num_highway = num_highway
+        self._output_size = output_size
+
+        with self.name_scope():
+            self._convs = nn.HybridSequential()
+            with self._convs.name_scope():
+                for _, ngram_size in enumerate(self._ngram_filter_sizes):
```
leezu: No need for enumerate

gluonnlp/model/char_encoder.py
```diff
@@ -0,0 +1,145 @@
...
+         booktitle={AAAI},
+         pages={2741--2749},
+         year={2016}
+        }
+
+    Parameters
+    ----------
+    embed_size : int
+        The input dimension to the encoder.
+    num_filters: int
```
leezu: It seems the original implementation sets `num_filters` as a function of `ngram_filter_sizes`. They use `min(200, 50*filter_size)`. Would it make sense to support setting different `num_filters`for different filter sizes?

gluonnlp/model/char_encoder.py
```diff
@@ -0,0 +1,145 @@
...
+    output_size: int
+        The output dimension after conducting the convolutions and max pooling,
+        and applying highways, as well as linear projection.
+
+    """
+    def __init__(self,
+                 embed_size,
+                 num_filters,
+                 ngram_filter_sizes,
+                 conv_layer_activation=nn.Activation('relu'),
```
leezu: It seems that the AAAI2016 paper differentiates between the activation function following convolution and the activation function following the highway layer. According to Table 2 they use tanh after Convolution and Relu after Highway. It seems here `conv_layer_activation` is also applied after the Highway layer?

gluonnlp/model/char_encoder.py
```diff
@@ -0,0 +1,145 @@
...
+            The output of the character encoder with shape `(batch_size, output_Size)`
+
+        """
+        if mask is not None:
+            inputs = inputs * mask.expand_dims(-1)
+
+        inputs = nd.transpose(inputs, axes=(1, 2, 0))
+
+        filter_outputs = []
+        for _, conv in enumerate(self._convs):
```
leezu: No need for enumerate

gluonnlp/model/char_encoder.py
```diff
@@ -0,0 +1,145 @@
...
+from mxnet.gluon import nn
+
+from .highway import Highway
+
+
+class CharacterEncoder(gluon.Block):
+    r"""Convolutional character encoder.
+
+    We implement the convolutional character encoder proposed in the following work::
+
```
leezu: Would it make sense to set the default parameters according to the paper?

gluonnlp/model/highway.py
```diff
@@ -0,0 +1,108 @@
...
+from mxnet import gluon
+from mxnet.gluon import nn
+
+
+class Highway(gluon.Block):
+    r"""Highway network.
+
+    We implemented the highway network proposed in the following work::
+
+        @article{srivastava2015highway,
```
leezu: It seems arXiv:1505.00387 is considered by the authors as extended abstract whereas _Srivastava, R. K., Greff, K., & Schmidhuber, J. (2015). Training very deep networks. In Advances in neural information processing systems (pp. 2377-2385)._ is the full paper. Should the latter be linked / cited as well?

gluonnlp/model/highway.py
leezu: Should this be `-1` (negative one)? Ie, does this initialize the bias of the transform gate? If so, it seems the following from the NIPS paper applies: 

> We use the transform gate defined as T(x) = σ(W_T^T x + b_T ) , where W_T is the weight matrix and b_T the bias vector for the transform gates.  This suggests a simple initialization scheme which is independent of the nature of H : b T can be initialized with a negative value (e.g. -1, -3 etc.) such that the network is initially biased towards carry behavior.

gluonnlp/model/highway.py
```diff
@@ -0,0 +1,108 @@
...
+                 **kwargs):
+        super(Highway, self).__init__(**kwargs)
+        self._input_size = input_size
+        self._num_layers = num_layers
+
+        with self.name_scope():
+            self.hnet = nn.HybridSequential()
+            with self.hnet.name_scope():
+                for _ in range(self._num_layers):
+                    self.hnet.add(nn.Dense(units=self._input_size * 2, in_units=self._input_size,
```
leezu: Would it make the code easier to read/maintain when using two separate Dense layers instead of a single one with twice the number of units and then splitting the units in half during forward? Then you can consider to use the `bias_initializer` parameter for the Dense layer that is used as Transform gate. It could be set by default to `mx.initializer.Constant(-1)` and users could overwrite it by passing an option like `transform_gate_bias_initializer` to `Highway`.

mli:
Job PR-129/17 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-129/17/index.html

docs/api/model.rst
cgraywang: Yes, the encoder could be useful for both character matrix as well as sentence matrix.

gluonnlp/model/highway.py
cgraywang: @szha based on offline discussion, we are using a highwaybias initializer now. Thanks 
@leezu Yes, after refering back to the paper, we are setting up the bias of the transform gate as -2 according to the AAAI paper. Thanks

gluonnlp/model/highway.py
cgraywang: Yes, changed to hybrid block.

gluonnlp/model/char_encoder.py
cgraywang: Thanks

gluonnlp/model/char_encoder.py
cgraywang: Now we allow users to specify the number of filters they want to apply according to different filter sizes. Thanks

gluonnlp/model/char_encoder.py
cgraywang: I originally thought the paper has the setting only on PTB dataset, so I didn't follow the setting. But after rethinking, Yes, I separate the activation functions.  The paper is using tanh for the conv and relu for the highway. Thanks for pointing out this.

gluonnlp/model/char_encoder.py
cgraywang: Yes.

gluonnlp/model/char_encoder.py
cgraywang: I originally thought the paper has the setting only on PTB dataset, so I didn't follow the setting.  Besides, I added the description in the docs.

gluonnlp/model/highway.py
cgraywang: Yes, let's cite both of them. Interestingly, the shorter one has more citations...

gluonnlp/model/highway.py
cgraywang: The `g` would be a non-linearity, the input projected and combined with the non-transform follows the definition of `g`. We are now using a highway bias to satisfy the requirement of highway initialization.

mli:
Job PR-129/19 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-129/19/index.html

docs/api/initializer.rst
```diff
@@ -0,0 +1,32 @@
+gluonnlp.initializer
+======================
+
+Gluon NLP Toolkit provides tools for easily customizing task specific initializer.
```
szha: update comment

mli:
Job PR-129/20 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-129/20/index.html

PR #134, opened 2018-06-05
Revert "Don't treat language model datasets special"

leezu:
Reverts dmlc/gluon-nlp#128

There is no consensus that usage of WikiText for non-LM purpose should be supported at the cost of making the LM API more complicated (by eg. making the `bptt_batchify` and `batchify` functions standalone).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -92,7 +92,7 @@ Load the Wikitext-2 dataset, for example:
 
     >>> import gluonnlp as nlp
     >>> train = nlp.data.WikiText2(segment='train')
-    >>> train[:5]
+    >>> train[0][0:5]
     ['=', 'Valkyria', 'Chronicles', 'III', '=']
 
 `Vocabulary Construction <http://gluon-nlp.mxnet.io/master/api/vocab.html>`__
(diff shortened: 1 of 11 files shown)


szha:
Thanks. I do think the general design of this refactor is a step in the right direction. This way, all the classes and interfaces are doing what they are supposed to, and WikiText classes will no longer be more specialized than what's necessary, which should open up possibility for more creative use and customization. Good design takes time, so thanks for proposing the change.

Let's review the design again after next release, and build consensus.

mli:
Job PR-134/2 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-134/2/index.html

mli:
Job PR-134/3 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-134/3/index.html

szha:
The merit of the design is not from the results it can get, but rather the flexibility and right principles for API design

cgraywang:
Yes, but I do think our package should balance between academic convention and flexible API usage

szha:
They are not in conflict.

PR #152, opened 2018-06-14
Add BiDAF embeddings

Ishitori:
## Description ##
Adds Char level embedding and Word level embedding layers needed for BiDAF model

### Essentials ###
- [ ] Code is well-documented

### Changes ###
- [ ] BiDAF's Char level embedding
- [ ] PredefinedEmbedding class - you can pass an embedding to be used as a layer
- [ ] BiDAFEmbedding that groups together both embeddings


mli:
Job PR-152/6 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-152/6/index.html

gluonnlp/model/block.py
```diff
@@ -89,3 +90,100 @@ def __init__(self, axis=-1, eps=1E-6, **kwargs):
...
+        # transposing to NCHW which is a default for Conv2D class
+        x_embedded_transposed = x_embedded.transpose(axes=(0, 3, 1, 2))
+        network_output = self.net.hybrid_forward(F, x_embedded_transposed)
+        relu_output = F.relu(network_output)
+        max_output = F.max(relu_output, axis=3)
+        # transposing back to expected batch_size x seq_len x channels
+        return max_output.transpose(axes=(0, 2, 1))
+
+
+class PredefinedEmbedding(HybridBlock):
```
szha: this type of block is only suitable for scripts.

gluonnlp/model/block.py
```diff
@@ -89,3 +90,100 @@ def __init__(self, axis=-1, eps=1E-6, **kwargs):
...
+        super(PredefinedEmbedding, self).__init__(prefix=prefix, params=params)
+        input_dim, output_dim = embedding.idx_to_vec.shape
+        self.embedding_data = embedding.idx_to_vec
+        self.embedding = nn.Embedding(input_dim, output_dim)
+
+    def hybrid_forward(self, F, x):  # pylint: disable=arguments-differ
+        return self.embedding.hybrid_forward(F, x, weight=self.embedding_data)
+
+
+class BiDAFEmbedding(HybridBlock):
```
szha: same, put in scripts. we can revisit whether to put this in API when we choose to provide pre-trained models.

mli:
Job PR-152/8 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-152/8/index.html

gluonnlp/model/block.py
```diff
@@ -89,3 +90,100 @@ def __init__(self, axis=-1, eps=1E-6, **kwargs):
     def hybrid_forward(self, F, x):  # pylint: disable=arguments-differ
         ret = F.broadcast_div(x, F.norm(x, axis=self._axis, keepdims=True) + self._eps)
         return ret
+
+
+class CharacterLevelCNNEmbedding(HybridBlock):
```
cgraywang: The CNN embedding used in the Bidaf paper follows this paper: https://arxiv.org/pdf/1508.06615.pdf, which uses highway and allows to specify multiple filters according different filter sizes. The input could be `(seq_len, batch_size, embedding_size)` and the output tensor should be `(batch_size, output_size)`, where the output_size could be the max_pooled_dimension. I already have the CNN embedding implementation in this PR: https://github.com/dmlc/gluon-nlp/pull/129. I suggested you could remove this character embedding in this PR.

gluonnlp/model/block.py
cgraywang: We expected a contextual embedding layer which leverages a LSTM on top of the character embedding and word embedding layer to model the interactions between words. This BiDAF embedding is not the contextual embedding and we don't need it a class.

gluonnlp/model/block.py
cgraywang: This embedding is not necessary since we can directly initialize a word embedding layer using the glove vectors: `net.embedding.weight.set_data(glove_embedding.idx_to_vec.as_in_context(ctx))`

PR #158, opened 2018-06-21
Add plain text of translation data.

szhengac:
## Description ##
This pr includes following feature:

1. add plain text of translation data 
2. fix dataset due to difference between src-ref and ref-src
3. update script to include the official evaluation of wmt

### Essentials ###
- [x] Code is well-documented

### Changes ###
- [] Feature1, tests, (and when applicable, API doc)
- [] Feature2, tests, (and when applicable, API doc)

## Comments ##

### DIFF
--- a/gluonnlp/data/transforms.py
+++ b/gluonnlp/data/transforms.py
@@ -23,7 +23,7 @@ clipping, padding, and tokenization."""
 from __future__ import absolute_import
 from __future__ import print_function
 
-__all__ = ['ClipSequence', 'PadSequence', 'NLTKMosesTokenizer', 'SpacyTokenizer']
+__all__ = ['ClipSequence', 'PadSequence', 'NLTKMosesTokenizer', 'SpacyTokenizer', 'NLTKMosesDetokenizer']
 
 import numpy as np
 import mxnet as mx
@@ -177,20 +177,23 @@ class NLTKMosesTokenizer(object):
                               'official installation guide in https://www.nltk.org/install.html .')
         self._tokenizer = MosesTokenizer()
 
-    def __call__(self, sample):
+    def __call__(self, sample, return_str=False):
         """
 
         Parameters
         ----------
         sample: str
             The sentence to tokenize
+        return_str: bool, default False
+            True: return a single string
+            False: return a list of tokens
 
         Returns
         -------
-        ret : list of strs
-            List of tokens
+        ret : list of strs or str
+            List of tokens or tokenized text
         """
-        return self._tokenizer.tokenize(sample)
+        return self._tokenizer.tokenize(sample, return_str=return_str)
 
 
 class SpacyTokenizer(object):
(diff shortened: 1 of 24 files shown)


gluonnlp/data/transforms.py
```diff
@@ -177,20 +178,23 @@ def __init__(self):
...
-    def __call__(self, sample):
+    def __call__(self, sample, return_str=False):
         """
 
         Parameters
         ----------
         sample: str
             The sentence to tokenize
+        return_str: bool, default False
+            True: return a single string
```
szha: why would a tokenizer want to return a single string? does some cleaning happen in this process?

szhengac: This can be used if we want to write the tokenized text into a file.

szhengac: In nltk, return_str=true means the function ignores str.split().

mli:
Job PR-158/3 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-158/3/index.html

gluonnlp/data/transforms.py
szha: This flag assumes the separator token to use when generating this string. I think doing this in a separate transformer function is likely cleaner. Is it easy to implement the joining logic to mimic the behavior in moses tokenizer?

szhengac: this corresponds to a single line in nltk tokenzier: `return text if return_str else text.split()`

szhengac: so they assume that separator is white space, which is hard decoded.

szhengac: The simple split() already removes more than one white spaces, why it should be different? The nltk implements the same logic as in Mose tokenizer, so I think their function is reasonable.

szha:
```
In [1]: from nltk.tokenize.moses import MosesTokenizer

In [2]: t = MosesTokenizer()

In [3]: x = ['a', ' ', ' ', 'b']

In [4]: xs = ''.join(x)

In [5]: xs
Out[5]: 'a  b'

In [6]: t.tokenize(xs)
Out[6]: ['a', 'b']

In [7]: t.tokenize(xs, return_str=True)
Out[7]: 'a b'
```

szha: note that the output is different from original string

szhengac: this is because Mose tokenizer will remove redundant space. So why it would be a problem? That's why we sometimes cannot recover the original string by calling tokenizer.perl and detokenizer.perl.

szha:
Please make sure that detokenizer is listed in the API doc

szha:
https://github.com/dmlc/gluon-nlp/blob/master/docs/api/data.rst#transforms

mli:
Job PR-158/4 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-158/4/index.html

mli:
Job PR-158/5 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-158/5/index.html

mli:
Job PR-158/13 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-158/13/index.html

scripts/nmt/machine_translation.rst
```diff
@@ -13,13 +13,22 @@ Use the following command to train the GNMT model on the IWSLT2015 dataset.
 
 It gets test BLEU score equals to 26.22.
 
-Use the following command to train the Transformer model on the WMT16 dataset for English to German translation. 
+Use the following commands to train the Transformer model on the WMT14 dataset for English to German translation.
 
 .. code-block:: console
```
StrayBird-ATSH, line 19: Hello, it seems that there is no reference about the paper in this section. Is it possible that you could add a reference so that it is easier for me to write the catalog page?
Thanks!
#1184

PR #186, opened 2018-07-02
Hybridize convolutional encoder

leezu:
## Description ##
This makes ConvolutionalEncoder a HybridBlock. It also adds a few more test cases.

### Essentials ###
- [X] Code is well-documented

### Changes ###
- [X] Allow hybridizing nlp.model.ConvolutionalEncoder

## Comments ##
- If the nlp.model.ConvolutionalEncoder is hybridized, and the first forward pass runs with/without passing a mask, subsequent passes must follow the same convention as otherwise the symbolic graph doesn't match. Users can create two instances of nlp.model.ConvolutionalEncoder with shared parameters if they require both masked and unmasked inputs at the same time. If the nlp.model.ConvolutionalEncoder is not hybridized there are no changes to the current behavior. However users may have a .hybridize() statement in their code which was a no-op currently and will take effect if this PR is merged.

### DIFF
--- a/gluonnlp/model/convolutional_encoder.py
+++ b/gluonnlp/model/convolutional_encoder.py
@@ -24,14 +24,14 @@ from __future__ import print_function
 
 __all__ = ['ConvolutionalEncoder']
 
-from mxnet import gluon, nd
+from mxnet import gluon
 from mxnet.gluon import nn
 from gluonnlp.initializer import HighwayBias
 
 from .highway import Highway
 
 
-class ConvolutionalEncoder(gluon.Block):
+class ConvolutionalEncoder(gluon.HybridBlock):
     r"""Convolutional encoder.
 
     We implement the convolutional encoder proposed in the following work::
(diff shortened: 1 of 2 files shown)


gluonnlp/model/convolutional_encoder.py
```diff
@@ -131,7 +133,13 @@ def __init__(self,
...
                 self._output_size = maxpool_output_size
 
-    def forward(self, inputs, mask=None): # pylint: disable=arguments-differ
+    def __call__(self, inputs, mask=None):
+        if mask is not None:
+            return super(ConvolutionalEncoder, self).__call__(inputs, mask)
+        else:
+            return super(ConvolutionalEncoder, self).__call__(inputs)
+
+    def hybrid_forward(self, F, inputs, mask=None): # pylint: disable=arguments-differ
```
szha: remove none default.

szha: then we can do away with the override of `__call__`

leezu: I don't think so, because for HybridBlock `None` is not a valid input: `E       AssertionError: HybridBlock input must be (nested) list of Symbol or NDArray, but got None of type <class 'NoneType'>`

leezu: Ie. to allow `encoder(inputs, None)` we must override __call__ or propose a change in mxnet if you believe this is actually not intended.

szha: Right, I met the exact same situation in gluon RNN layer. The trick is to use positional arguments for hybrid_forward such as `hybrid_forward(self, F, *inputs)`

leezu: Even with positional arguments I get`E AssertionError: HybridBlock input must be (nested) list of Symbol or NDArray, but got None of type <class 'NoneType'>`. The reason I overwrote `__call__` is to preserve the current behavior where users can set mask explicitly to None, ie. call: `encoder(inputs, None)` compared to `encoder(inputs)` where the mask just defaults to None. If you are fine with only supporting the latter we can remove the `__call__` overwrite.
`encoder(inputs, mask)` with mask being a proper NDArray is always possible.

szha: We can drop the support for encoder(inputs, None).

mli:
Job PR-186/10 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-186/10/index.html

PR #193, opened 2018-07-03
Remove gluon.utils.download copy with verify_ssl support

leezu:
## Description ##
The problematic server now servers a valid SSL certificate. Also https://github.com/apache/incubator-mxnet/pull/11546 will add the option upstream.

### Essentials ###
- [X] Code is well-documented

### DIFF
--- a/env/doc.yml
+++ b/env/doc.yml
@@ -14,5 +14,5 @@ dependencies:
   - nltk
   - pandoc==2.2.1
   - pip:
-    - mxnet-cu80>=1.3.0b20180629
+    - mxnet-cu80>=1.3.0b20180725
     - notedown
--- a/env/py2.yml
+++ b/env/py2.yml
@@ -11,4 +11,4 @@ dependencies:
   - flaky
   - pytest-cov
   - pip:
-    - mxnet-cu80>=1.3.0b20180629
+    - mxnet-cu80>=1.3.0b20180725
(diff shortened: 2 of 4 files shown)


szha:
We will need to wait until we update the dependency with a nightly version that includes the update in download before we merge.

leezu:
Currently the updated `download` is not needed any more, because the third-party server was fixed. I proposed the change in mxnet as we may run into similar issues in future.

leezu:
It turns out the webserver is still not fixed. The issue is that it does not serve an intermediary certificate that is needed to establish the trust from the root CA to the websites certificate. Newer browsers include the intermediary certificate used, so no error is shown when accessing the file with Firefox.

I have contacted the admin and asked to include the intermediary certificates.  More info about the certificate issues can be found here https://www.ssllabs.com/ssltest/analyze.html?d=ie.technion.ac.il

mli:
Job PR-193/16 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-193/16/index.html

szha:
@leezu it passed. Was the issue upstream fixed?

leezu:
No, I made use of the verify_ssl=False that was added to upstream mxnet.

leezu:
I haven't received a response from the webmaster. They seem unwilling to fix their server.

PR #201, opened 2018-07-11
attention assert

hhexiy:
## Description ##
Luong style attention should make sure that the projected query has the same dimension as the key, otherwise batch_dot will have dimension mismatch.

### Changes ###
- add dimension assert for Luong style attention

### DIFF
--- a/gluonnlp/model/attention_cell.py
+++ b/gluonnlp/model/attention_cell.py
@@ -457,8 +457,8 @@ class DotProductAttentionCell(AttentionCell):
             if not self._luong_style:
                 key = self._proj_key(key)
             else:
-                assert query.shape[-1] == key.shape[-1], 'Luong style attention requires key to have ' \
-                                                         'the same dim as the projected query'
+                assert query.shape[-1] == key.shape[-1], 'Luong style attention requires key to ' \
+                                                         'have the same dim as the projected query'
         if self._normalized:
             query = self._l2_norm(query)
             key = self._l2_norm(key)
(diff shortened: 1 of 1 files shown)


gluonnlp/model/attention_cell.py
```diff
@@ -457,8 +457,8 @@ def _compute_weight(self, F, query, key, mask=None):
             if not self._luong_style:
                 key = self._proj_key(key)
             else:
-                assert query.shape[-1] == key.shape[-1], 'Luong style attention requires key to have ' \
-                                                         'the same dim as the projected query'
+                assert query.shape[-1] == key.shape[-1], 'Luong style attention requires key to ' \
+                                                         'have the same dim as the projected query'
```
szha: could you include the shapes in the error message?

gluonnlp/model/attention_cell.py
```diff
@@ -456,6 +456,11 @@ def _compute_weight(self, F, query, key, mask=None):
             query = self._proj_query(query)
             if not self._luong_style:
                 key = self._proj_key(key)
+            else:
+                assert query.shape[-1] == key.shape[-1], 'Luong style attention requires key to ' \
```
szha: query.shape is only available if query is ndarray. you can do the check by checking whether `F == nd`.

hhexiy: added.

mli:
Job PR-201/5 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-201/5/index.html

PR #209, opened 2018-07-15
Fix off by one error for ConstWidthBucket

leezu:
``` python
list(nlp.data.FixedBucketSampler(
    np.arange(16, 32), 8, num_buckets=2
    bucket_scheme=nlp.data.ConstWidthBucket()))
```

Should result in exactly two batches of batch_size 8:
``` python
[[8, 9, 10, 11, 12, 13, 14, 15], [0, 1, 2, 3, 4, 5, 6, 7]]
```
Without this change, 3 batches with batch_size varying between 1, 7 and 8 are
generated.

### Essentials ###
- [X] Code is well-documented

### Changes ###
- [X] Fix off by one error for ConstWidthBucket

## Comments ##
@szhengac please let me know if this has any negative impact on your use-case.

### DIFF
--- a/gluonnlp/data/sampler.py
+++ b/gluonnlp/data/sampler.py
@@ -123,7 +123,7 @@ class ConstWidthBucket(BucketScheme):
             A list including the keys of the buckets.
         """
         if not isinstance(max_lengths, INT_TYPES):
-            bucket_width_l = [max((max_len - min_len) // num_buckets, 1)
+            bucket_width_l = [max((1 + max_len - min_len) // num_buckets, 1)
                               for max_len, min_len in
                               zip(max_lengths, min_lengths)]
             bucket_keys = \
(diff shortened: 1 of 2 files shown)


szha:
@leezu seems like it requires a rebase.

szhengac:
The example seems problematic, as the input length varies from 16 to 31,  but you have output from 0 to 15. I see your point, but we may still have similar results if you use `[np.arange(16, 32)] + [16]`, then the output will have batch size varying betwee 1 and 8.

mli:
Job PR-209/2 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-209/2/index.html

szha:
@leezu would it be easy to add a test case?

mli:
Job PR-209/5 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-209/5/index.html

PR #215, opened 2018-07-17
Use feature detection instead of version detection

cclauss:
Follows the Python porting best practice [use feature detection instead of version detection](https://docs.python.org/3/howto/pyporting.html#use-feature-detection-instead-of-version-detection).

### DIFF
--- a/gluonnlp/base.py
+++ b/gluonnlp/base.py
@@ -22,9 +22,7 @@
 
 __all__ = ['_str_types']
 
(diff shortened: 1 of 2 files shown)


mli:
Job PR-215/2 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-215/2/index.html

PR #216, opened 2018-07-17
add dtype to batchify function

szhengac:
## Description ##
add dtype to batchify function

### Essentials ###
- [x] Code is well-documented

### Changes ###

## Comments ##

### DIFF
--- a/gluonnlp/data/batchify.py
+++ b/gluonnlp/data/batchify.py
@@ -24,7 +24,7 @@ import numpy as np
 import mxnet as mx
 
 
-def _pad_arrs_to_max_length(arrs, pad_axis, pad_val, use_shared_mem=False):
+def _pad_arrs_to_max_length(arrs, pad_axis, pad_val, use_shared_mem=False, dtype=None):
     """Inner Implementation of the Pad batchify
 
     Parameters
@@ -47,12 +47,19 @@ def _pad_arrs_to_max_length(arrs, pad_axis, pad_val, use_shared_mem=False):
     ret_shape[pad_axis] = max_size
     ret_shape = (len(arrs), ) + tuple(ret_shape)
     if use_shared_mem:
-        ret = mx.nd.full(shape=ret_shape, val=pad_val, ctx=mx.Context('cpu_shared', 0),
-                         dtype=arrs[0].dtype)
+        if dtype is not None:
+            ret = mx.nd.full(shape=ret_shape, val=pad_val, ctx=mx.Context('cpu_shared', 0),
+                             dtype=dtype)
+        else:
+            ret = mx.nd.full(shape=ret_shape, val=pad_val, ctx=mx.Context('cpu_shared', 0),
+                             dtype=arrs[0].dtype)
         original_length = mx.nd.array(original_length, ctx=mx.Context('cpu_shared', 0),
                                       dtype=np.int32)
     else:
-        ret = mx.nd.full(shape=ret_shape, val=pad_val, dtype=arrs[0].dtype)
+        if dtype is not None:
+            ret = mx.nd.full(shape=ret_shape, val=pad_val, dtype=dtype)
+        else:
+            ret = mx.nd.full(shape=ret_shape, val=pad_val, dtype=arrs[0].dtype)
         original_length = mx.nd.array(original_length, dtype=np.int32)
     for i, arr in enumerate(arrs):
         if arr.shape[pad_axis] == max_size:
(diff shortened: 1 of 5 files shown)


gluonnlp/data/batchify.py
```diff
@@ -47,12 +47,19 @@ def _pad_arrs_to_max_length(arrs, pad_axis, pad_val, use_shared_mem=False):
...
     ret_shape = (len(arrs), ) + tuple(ret_shape)
     if use_shared_mem:
-        ret = mx.nd.full(shape=ret_shape, val=pad_val, ctx=mx.Context('cpu_shared', 0),
-                         dtype=arrs[0].dtype)
+        if dtype is not None:
+            ret = mx.nd.full(shape=ret_shape, val=pad_val, ctx=mx.Context('cpu_shared', 0),
+                             dtype=dtype)
+        else:
+            ret = mx.nd.full(shape=ret_shape, val=pad_val, ctx=mx.Context('cpu_shared', 0),
+                             dtype=arrs[0].dtype)
```
szha:
```python
dtype = arrs[0].dtype if dtype is None else dtype
ret = mx.nd.full(shape=ret_shape, val=pad_val, ctx=mx.Context('cpu_shared', 0),
                           dtype=dtype)
```

tests/unittest/test_batchify.py
```diff
@@ -5,18 +5,21 @@
...
-        batchify_out = batchify_fn(dat).asnumpy()
-        npy_out = np.array(dat)
-        assert_allclose(batchify_out, npy_out)
-        assert batchify_out.dtype == npy_out.dtype
+        for _dtype in [None, dtype]:
+            batchify_fn = batchify.Stack(dtype=_dtype)
+            batchify_out = batchify_fn(dat).asnumpy()
+            npy_out = np.array(dat)
+            assert_allclose(batchify_out, npy_out)
+            assert batchify_out.dtype == npy_out.dtype
```
sxjscience: It has not checked the case that the `dtype` used in Stack is different from the `dtype` of the input array. We may do the following instead:
```python
for _dtype in [np.float32]:
     batchify_fn = batchify.Stack(dtype=_dtype)
     batchify_out = batchify_fn(dat).asnumpy()
     assert batchify_out.dtype == np.float32
```

gluonnlp/data/batchify.py
```diff
@@ -66,23 +67,38 @@ def _pad_arrs_to_max_length(arrs, pad_axis, pad_val, use_shared_mem=False):
...
         else:
             return mx.nd.stack(*arrs)
     else:
         out = np.asarray(arrs)
+        dtype = out.dtype if dtype is None else dtype
         if use_shared_mem:
-            return mx.nd.array(out, ctx=mx.Context('cpu_shared', 0), dtype=out.dtype)
+            return mx.nd.array(out, ctx=mx.Context('cpu_shared', 0), dtype=dtype)
         else:
             return mx.nd.array(out, dtype=out.dtype)
```
sxjscience: Should be `return mx.nd.array(out, dtype=dtype)`.

tests/unittest/test_batchify.py
```diff
@@ -37,62 +40,69 @@ def _verify_padded_arr(padded_arr, original_arr, pad_axis, pad_val, pad_length,
...
-                                                    ret_use_mx[i][1].asnumpy())
-                                    assert(ret_use_npy[i][1].shape == (batch_size,))
-                                else:
-                                    assert ret_use_npy[i].dtype == ret_use_mx[i].dtype == dtype
-                                    assert_allclose(ret_use_npy[i].asnumpy(), ret_use_mx[i].asnumpy())
+                        for _dtype in [None, dtype]:
+                            shapes = [[2 for _ in range(ndim)] for _ in range(batch_size)]
+                            for i in range(len(shapes)):
+                                shapes[i][axis] = np.random.randint(length_min, length_max)
+                            random_data_npy = [np.random.normal(0, 1, shape).astype(dtype)
```
sxjscience: Since we set the random_data_npy to have `dtype`, the check also does not test the case that the `dtypes` are different.

gluonnlp/data/batchify.py
```diff
@@ -66,23 +67,38 @@ def _pad_arrs_to_max_length(arrs, pad_axis, pad_val, use_shared_mem=False):
...
+    r""""Base class for applying batchify function on input data.
+
+    Parameters
+    ----------
+    dtype : str or numpy.dtype, default None
+        The value type of the output. If it is set to None, the input data type is used.
+    """
+
+    def __init__(self, dtype=None):
+        self._dtype = dtype
```
sxjscience: I suggest manually adding the `dtype` argument to `Stack` and `Pad`. This base class looks redundant.

szhengac: This may be cleaner if we add another batchify function in the future?

sxjscience: Since it has only one argument, it would be okay to add it manually to other classes. Just a single-line of code. Also, `Tuple` cannot be inherit `Batchify`.

szhengac: Do we really need to have Tuple inherit Batchify? Tuple is only supposed to wrap other batchify functions.

sxjscience: Not really. The main concern is that the base class has just one argument and is equivalent to a single-line of code. I think it would be better to keep the interface as simple as possible and avoid adding too many hierarchies.

szhengac: Okay

szhengac: But we have duplicate doc.

mli:
Job PR-216/3 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-216/3/index.html

mli:
Job PR-216/5 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-216/5/index.html

PR #217, opened 2018-07-18
Fix UnigramCandidateSampler

leezu:
## Description ##
This fixes `UnigramCandidateSampler`.

### Essentials ###
- [X] Code is well-documented

### Changes ###
- [X] Fix `UnigramCandidateSampler`

## Comments ##
- `dtype='float64'` seems sufficient to avoid numerical issues (meaning out of range samples) for up to [0, 9M]. With float32 for values as small as 100k numerical issues would result in frequently sampling a value that is one too high. Due to https://github.com/apache/incubator-mxnet/issues/11551 this would go unnoticed and just sometimes lead to obscure crashes depending on the memory value read at the out of bound location.
- Also high should be == N, not N+1.

### DIFF
--- a/gluonnlp/data/candidate_sampler.py
+++ b/gluonnlp/data/candidate_sampler.py
@@ -100,8 +100,9 @@ class UnigramCandidateSampler(CandidateSampler):
         samples: NDArray
             The sampled candidate classes.
         """
-        idx = mx.nd.random.uniform(low=0, high=self.N + 1, shape=shape,
-                                   ctx=self._context).floor()
+        idx = mx.nd.random.uniform(low=0, high=self.N, shape=shape,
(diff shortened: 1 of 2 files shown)


szhengac:
Have you tried numpy? Would it have the same issue?

leezu:
We can use `np.random.randint`, but ideally we want to avoid it as it is a blocking call (ie. delaying further excution of the Python program), whereas `mx.nd.random.uniform` just pushes the job of creating a random array to the Engine. It would be best to implement `mx.nd.random.randint`.

mli:
Job PR-217/4 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-217/4/index.html

PR #228, opened 2018-07-24
Re-initialize the unknown token vector for npz distributed on S3 by default

leezu:
## Description ##
We currently distribute some pre-trained embeddings on with pre-initialized unknown token. This is known and a workaround has been implemented. Currently the workaround treats the unknown token and vector loaded from the npz file as correct, even if the user specifies a special initialization function, instead of re-initializing it.

The reasoning was, that for some embeddings the unknown token vector in the npz file may actually be valuable and shouldn't be overwritten. However, this was not a good default behavior. With this PR, the unknown vector from the npz will only be preserved if `init_unknown_vec=None`.
Note this only affects embeddings distributed by us on S3.

Thanks @garima3292 for reporting this issue.

### Essentials ###
- [X] Code is well-documented

### Changes ###
- [X] Fix `init_unknown_vec` ignored for some pretrained embeddings.

### DIFF
--- a/gluonnlp/embedding/token_embedding.py
+++ b/gluonnlp/embedding/token_embedding.py
@@ -341,19 +341,13 @@ class TokenEmbedding(object):
             # is the same now as it was when the .npz was generated. Under this
             # assumption we can safely overwrite the respective token and
             # vector from the npz.
-            if deserialized_embedding.unknown_token == self.unknown_token:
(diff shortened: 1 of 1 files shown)


mli:
Job PR-228/5 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-228/5/index.html

PR #218, opened 2018-07-18
Add Wikipedia dump streams and ContextStream

leezu:
## Description ##
This adds Wikipedia Corpus Stream for embedding training script based on #235.

### Essentials ###
- [x] Changes are complete (i.e. I finished coding on this PR)
- [X] All changes have test coverage
- [X] Code is well-documented

### Changes ###
- [X] ContextStream to transform a CorpusStream into a stream of batches for embedding training
- [X] Restructure train_fasttext.py example to make use of streams
- [X] Add WikiDumpStream to scripts folder

PR #235, opened 2018-07-29
Refactor DataStream

leezu:
## Description ##
See Changes below. This was part of https://github.com/dmlc/gluon-nlp/pull/218. I split up the PR to make it easier to review / discuss.

### Essentials ###
- [X] Code is well-documented

### Changes ###
- [X] Remove DataStream.transform() and encourage the use of Python generator expressions instead
- [X] Remove SimpleDataStream (without .transform(), any Python iterable implements the DataStream interface. Consequently there is no need for SimpleDataStream. Let's keep the class hierarchy simple.)
- [X] Refactor PrefetchingStream
  - Remove option to iterate over a list of Streams. This allows keeping the implementation simple. A list of streams can be transformed into a single stream using `zip`.
  - Add option to use multi-processing (with a single worker process)
- [X] Introduce DatasetStream to clarify when we expect a Stream over Datasets vs over individual samples
- [X] Refactor CorpusStream to iterate over CorpusDataset instead of single sentences
  - An iterator over single sentences can still be obtained via `itertools.chain.from_iterable(corpus_stream)`. However, it can be useful to iterate over bigger chunks in the stream to speed up other steps in the data pipeline.
  - Remove the `sampler` argument. It is not needed anymore. (Note that it still kept for LanguageModelStream, which internally now uses `itertools.chain.from_iterable(corpus_stream)`)
- [X] Fix random stability of PrefetchingStream
  - Set the python, numpy and mxnet seeds in the started thread / process deterministically based on the python random number generator state in the calling thread.

## Comments ##
- We may want to re-consider introducing `DatasetStream`. The only difference to `DataStream` is the docstring. It may or may not be a good idea to complicate the class hierarchy for this.

### DIFF
--- a/gluonnlp/data/stream.py
+++ b/gluonnlp/data/stream.py
@@ -56,6 +56,17 @@ class DataStream(object):
     """
 
     def __iter__(self):
+        """Return an iterator over all elements of the DataStream.
+
+        This method returns a new iterator object that can iterate over
+        all the objects in the DataStream.
+
+        Returns
+        -------
+        iterator
+            An object implementing the Python *iterator protocol*.
+
+        """
         raise NotImplementedError
 
     def transform(self, fn):
--- a/gluonnlp/data/stream.py
+++ b/gluonnlp/data/stream.py
@@ -24,18 +24,18 @@ ready for training and evaluation."""
 __all__ = ['DataStream', 'CorpusStream', 'LanguageModelStream', 'SimpleDataStream',
            'PrefetchingStream']
 
-import os
 import glob
 import itertools
-import threading
-import numpy as np
+import multiprocessing
 import os
 import random
-import multiprocessing
 import threading
+
+import numpy as np
 
 import mxnet as mx
 from mxnet.gluon.data import RandomSampler, SequentialSampler
+
 from .dataset import CorpusDataset
 
 try:
--- a/gluonnlp/data/stream.py
+++ b/gluonnlp/data/stream.py
@@ -485,7 +485,7 @@ class _ThreadPrefetcher(_Prefetcher, threading.Thread):
         self.start()
 
 
-class PrefetchingStream(object):
+class PrefetchingStream(DataStream):
     """Prefetch a DataStream in a separate Thread or Process.
 
     This iterator will create another thread or process to perform
--- a/gluonnlp/data/stream.py
+++ b/gluonnlp/data/stream.py
@@ -415,7 +415,7 @@ class _LanguageModelBPTTStream(DataStream):
 
 
 class _Prefetcher(object):
-    """Interal shared prefetcher logic."""
+    """Internal shared prefetcher logic."""
     data_queue = None
     control_queue = None
 
@@ -464,7 +464,7 @@ class _Prefetcher(object):
 
 
 class _ProcessPrefetcher(_Prefetcher, multiprocessing.Process):
-    """Interal multi-processing prefetcher."""
+    """Internal multi-processing prefetcher."""
 
     def __init__(self, *args, **kwargs):
         super(_ProcessPrefetcher, self).__init__(*args, **kwargs)
@@ -475,7 +475,7 @@ class _ProcessPrefetcher(_Prefetcher, multiprocessing.Process):
 
 
 class _ThreadPrefetcher(_Prefetcher, threading.Thread):
-    """Interal threaded prefetcher."""
+    """Internal threaded prefetcher."""
 
     def __init__(self, *args, **kwargs):
         super(_ThreadPrefetcher, self).__init__(*args, **kwargs)
--- a/gluonnlp/data/stream.py
+++ b/gluonnlp/data/stream.py
@@ -407,14 +407,21 @@ class _Prefetcher(object):
     data_queue = None
     control_queue = None
 
-    def __init__(self, stream, num_prefetch):
+    def __init__(self, stream, num_prefetch, seed, np_seed, mx_seed):
         super(_Prefetcher, self).__init__()
         self.stream = stream
         assert num_prefetch > 0, 'Unbounded Prefetcher is unsupported.'
         self.num_prefetch = num_prefetch
+        self.seed = seed
+        self.np_seed = np_seed
+        self.mx_seed = mx_seed
 
     def run(self):
         """Method representing the process’s activity."""
+        random.seed(self.seed)
+        np.random.seed(self.np_seed)
+        mx.random.seed(self.mx_seed)
+
         stream_iter = iter(self.stream)
         while True:
             try:  # Check control queue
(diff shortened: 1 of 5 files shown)


leezu:
With this change, we should also define exactly what we consider to be a stream (or a valid input to a `SimpleDataStream`). Do we consider iterable objects (has an `__iter__`method which returns an iterator; or defines `__getitem__`) or iterator objects (has a `next` (Python 2) or `__next__ `(Python 3) method.) as valid input? The main difference is that an if we call `iter(iterable)` multiple times, we will get independent iterators, whereas we can only iterate once over a single iterator.

In the context of GluonNLP DataStream this means, do we want to allow iterating over a DataStream multiple times? Or do we consider a DataStream to be a throwaway object that should only be used once? (It is cheap to create a new DataStream).

This PR proposed using generator expression `transformed = (transform(s) for s in stream)` instead of of `stream.transform()`. Note that the generator expression always returns a iterator. `stream.transform()` however always returns a iterable, meaning that if the underlying stream is an iterable it is possible to iterate multiple times over the transformed stream (and in case the underlying stream is an iterator, it is only possible to iterate once over the transformed stream).

Consequently, using generator expressions only makes sense if we only require the stream to be an iterator. If we decide to require the Stream to be an iterable I will revert the respective change and add back the `.transform()` interface.

(Independently of this question, in the previous implementation `PrefetchingStream(stream).transform()` wouldn't work.)

A small example to illustrate the difference:

```
In [1]: import gluonnlp as nlp
In [2]: iterator = [1,2,3]
In [3]: iterable = (i for i in range(3))
In [4]: stream1 = nlp.data.SimpleDataStream(iterator)
In [5]: stream2 = nlp.data.SimpleDataStream(iterable)
In [6]: list(stream1)
Out[6]: [1, 2, 3]
In [7]: list(stream1)
Out[7]: [1, 2, 3]
In [8]: list(stream2)
Out[8]: [0, 1, 2]
In [9]: list(stream2)
Out[9]: []
```

gluonnlp/data/stream.py
```diff
@@ -382,20 +312,23 @@ class _LanguageModelBPTTStream(DataStream):
         The length of each of the samples for truncated back-propagation-through-time (TBPTT).
     batch_size : int
         The number of samples in each batch.
+     sampler : mx.gluon.data.Sampler
```
eric-haibin-lin: nit: indentation

gluonnlp/data/stream.py
```diff
@@ -453,3 +388,117 @@ def _write(data, target, buffers, seq_len, i, length):
...
+    will be initialized randomly based on the next 32 bit integer in the
+    python, numpy and mxnet random generator of the caller respectively
+    (random.getrandbits(32), numpy.random.randint(0, 2**32),
+    int(mx.nd.random.uniform(0, 2**32).asscalar())).
+
+    Parameters
+    ----------
+    stream : DataStream
+        Source stream.
+    num_prefetch : int, default 1
```
eric-haibin-lin: is `num_prefetch ` > 1 also supported?

leezu: It should be. num_prefetch sets the size of the queue. The Thread / Process tries to push as many elementsas possible to the queue.

mli:
Job PR-235/15 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-235/15/index.html

gluonnlp/data/stream.py
leezu: I updated the tests to run with different values of num_prefetch as well as thread / process.

gluonnlp/data/stream.py
leezu: thanks! fixed

mli:
Job PR-235/16 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-235/16/index.html

mli:
Job PR-235/25 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-235/25/index.html

mli:
Job PR-235/26 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-235/26/index.html

mli:
Job PR-235/34 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-235/34/index.html

mli:
Job PR-235/36 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-235/36/index.html

gluonnlp/data/stream.py
```diff
@@ -453,3 +400,119 @@ def _write(data, target, buffers, seq_len, i, length):
             if has_token_buffered or self._last_batch == 'keep':
                 yield mx.nd.array(data).T, mx.nd.array(target).T, mx.nd.array(mask).T
         return
+
+
+class _Prefetcher(object):
+    """Interal shared prefetcher logic."""
```
eric-haibin-lin: nit: Interal -> Internal

gluonnlp/data/stream.py
```diff
@@ -453,3 +400,119 @@ def _write(data, target, buffers, seq_len, i, length):
...
+
+    def next(self):
+        return self.__next__()
+
+    def __iter__(self):
+        return self
+
+
+class _ProcessPrefetcher(_Prefetcher, multiprocessing.Process):
+    """Interal multi-processing prefetcher."""
```
eric-haibin-lin: Internal

gluonnlp/data/stream.py
```diff
@@ -453,3 +400,119 @@ def _write(data, target, buffers, seq_len, i, length):
...
+    def __init__(self, *args, **kwargs):
+        super(_ProcessPrefetcher, self).__init__(*args, **kwargs)
+        self.data_queue = multiprocessing.Queue(self.num_prefetch)
+        self.control_queue = multiprocessing.Queue()
+        self.daemon = True
+        self.start()
+
+
+class _ThreadPrefetcher(_Prefetcher, threading.Thread):
+    """Interal threaded prefetcher."""
```
eric-haibin-lin: Internal

gluonnlp/data/stream.py
eric-haibin-lin: Oh okay I was under the wrong impression that this is the number of prefetching threads. nvm

mli:
Job PR-235/38 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-235/38/index.html

PR #237, opened 2018-07-31
Add dataloader that allows multi-shard sampling

szhengac:
## Description ##
The current dataloader and split_and_load function can consume a lot of unnecessary memory for paddings with multi-gpu training. Imagine that we want to sample a large mini-batch and split it into multiple gpus. With default dataloader and split_and_load function, we will first perform paddings on the sequences and then split the formed matrix into num_gpus smaller ones, and therefore the number of paddings is determined by the longest sequence across the gpus, while such paddings may only be needed in one particular gpu. The amount of paddings can be much larger if we do not clip the data sequences and the lengths in the buckets have much larger variance. 

So in this pr, I propose to sample num_gpus smaller mini-batches at each iteration, and each small mini-batch is assigned to each gpu. In this case, the paddings for each small samples is performed independently. To achieve such goal, we only require sampler and dataloader to sample/receive list(list(int)) data, respectively.

The previous implementation does not do well in the AWS P3.16 instance (8 GPUS), in particular, the training is significantly slower than in the AWS P3.8 instance (4 GPUS). The experiments show that, with the proposed dataloader, we can use about 2x times larger mini-batch, and the training is faster in AWS P3.16 instance compared to in the AWS P3.8 instance. The current implementation in master requires 57 hours to reach test bleu score (bleu -13a) 26.89 in full newstest2014 dataset in AWS P3.8 instance, with the proposed change, it only requires 24 hours to reach test bleu score (bleu -13a) 27.05 in the AWS P3.16 instance. The improvement of bleu score is due to the use of larger mini-batches and therefore the larger learning rate. 

### Essentials ###
- [x] Code is well-documented

### Changes ###
- add ShardedDataLoader
- make FixedBucketSample can output list(list(int)) batches

## Comments ##


mli:
Job PR-237/3 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-237/3/index.html

mli:
Job PR-237/5 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-237/5/index.html

mli:
Job PR-237/10 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-237/10/index.html

PR #243, opened 2018-08-01
Fix LanguageModelDataset.bptt_batchify

leezu:
## Description ##
In the current `LanguageModelDataset.bptt_batchify`, elements of the "last batch" (defined as the extra words at the end of the dataset: `data[(len(data) // batch_size) * batch_size:]`) would incorrectly be split evenly among all rows in the batch dimension. However, each row in our batches should contain a independent context. The correct context for the words in the "last batch" is only the very last row. This PR fixes the logic and adds tests.

### Essentials ###
- [X] Code is well-documented

### Changes ###
- [X] Fix last batch in LanguageModelDataset.bptt_batchify

### DIFF
--- a/gluonnlp/data/dataset.py
+++ b/gluonnlp/data/dataset.py
@@ -24,6 +24,7 @@ __all__ = ['TextLineDataset', 'CorpusDataset', 'LanguageModelDataset']
 
 import io
 import os
+import math
 
 import mxnet as mx
 from mxnet.gluon.data import SimpleDataset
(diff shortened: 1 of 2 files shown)


gluonnlp/data/dataset.py
```diff
@@ -195,19 +196,24 @@ def bptt_batchify(self, vocab, seq_len, batch_size, last_batch='keep'):
...
-        batches = slice_sequence(data, seq_len+1, overlap=1)
+        if last_batch not in ['keep', 'discard']:
+            raise ValueError('last_batch must be "keep" or "discard".')
+
         if last_batch == 'keep':
+            if not vocab.padding_token:
+                raise ValueError('vocab.padding_token must be specified '
+                                 'in vocab when last_batch="keep".')
+            coded = vocab[self._data[0]]
+            sample_len = math.ceil(len(coded) / batch_size)
```
szha: in py2 the division results in integer

leezu: Thanks

leezu:
@szha @cgraywang why the strict inequality here: https://github.com/dmlc/gluon-nlp/blame/master/tests/unittest/test_datasets.py#L114

leezu:
The current implementation in this PR results in

```
        train_data = train.bptt_batchify(vocab, 35, 80, last_batch='keep')

        assert len(train_data) == 742, len(train_data)

>       assert train_data[-1][0].shape[0] < 35

E       assert 35L < 35



```

szha:
That test was intended for testing the case where last sample is shorter.

leezu:
Yes. But the current logic in utils.py will always pad such that seq_len is reached (ie. 35 in this case). Would there be any issue with changing the behavior (as in this PR) so that all batches always have seq_len * batch_size elements?

szha:
shouldn't be any issue as long as the language modeling scripts are correctly handling it. the change makes sense with the mask being introduced.

cc'd @cgraywang

mli:
Job PR-243/5 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-243/5/index.html

leezu:
Actually I believe the behavior implemented now in this PR is the only correct way to handle padding without discarding any data while still making sure the batch indices in the last batch are correct continuations of the previous batches. Ie. the previous test was asserting the incorrect behavior.

Also, the fix doesn't affect masking. Nevertheless we may want to add masking to the word_language_model.py script to make sure the perplexity calculation is correct (not affected by padded entries).

gluonnlp/data/dataset.py
```diff
@@ -195,19 +196,24 @@ def bptt_batchify(self, vocab, seq_len, batch_size, last_batch='keep'):
 
             - discard: The last batch is discarded if it's smaller than `(seq_len, batch_size)`.
         """
-        data = self.batchify(vocab, batch_size)
-        batches = slice_sequence(data, seq_len+1, overlap=1)
+        if last_batch not in ['keep', 'discard']:
+            raise ValueError('last_batch must be "keep" or "discard".')
```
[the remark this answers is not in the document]
leezu: Thanks. Fixed it and rebased

mli:
Job PR-243/7 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-243/7/index.html

leezu:
@szha FYI, seems #248 did not fix the non-deterministic CI failures. In http://ci.mxnet.io/blue/organizations/jenkins/gluon-nlp/detail/PR-243/7/pipeline the Deploy stage ran twice due to the JSONDecodError.

PR #249, opened 2018-08-02
Speed up batchify.Pad

leezu:
## Description ##
batchify.Pad currently allocates a large NDArray of the final shape, then converts all to be Padded datastructures to mx.nd.NDArray and then writes each axis of the large NDArray with the to be Padded array (thereby padding it). This is highly undesirable as repeatedly writing to the different axes of the NDArray is very expensive. This PR changes the internal implementation to convert everything to numpy instead which results in a 50x Speedup (even when the input is a list of ndarrays).

### Essentials ###
- [X] Code is well-documented

### Changes ###
- [X] Speed up gluonnlp.data.batchify.Pad

## Comments ##
This leads to a 50 times speed up for all supported input datastructures (nested python lists, numpy arrays, ndarrays) according to the following benchmark script:

```
import mxnet as mx
import random
import numpy as np
import gluonnlp as nlp

random.seed(1)
p = nlp.data.batchify.Pad()

# l = [np.array([1] * random.randint(5, 50)) for _ in range(500)]
# l = [mx.nd.array([1] * random.randint(5, 50)) for _ in range(500)]
l = [[1] * random.randint(5, 50) for _ in range(500)]

for _ in range(1000):
    p(l)

```

### DIFF
--- a/docs/api/data.batchify.rst
+++ b/docs/api/data.batchify.rst
@@ -22,3 +22,4 @@ API Reference
 .. automodule:: gluonnlp.data.batchify
    :members:
    :imported-members:
+   :special-members: __call__
--- a/gluonnlp/data/batchify.py
+++ b/gluonnlp/data/batchify.py
@@ -20,6 +20,8 @@
 into batches for fast processing."""
 __all__ = ['Stack', 'Pad', 'Tuple']
 
+import logging
+
 import numpy as np
 import mxnet as mx
 
@@ -39,22 +41,23 @@ def _pad_arrs_to_max_length(arrs, pad_axis, pad_val, use_shared_mem, dtype):
     ret : NDArray
     original_length : NDArray
     """
-    if not isinstance(arrs[0], (mx.nd.NDArray, np.ndarray)):
+    if isinstance(arrs[0], mx.nd.NDArray):
+        dtype = arrs[0].dtype if dtype is None else dtype
+        arrs = [arr.asnumpy() for arr in arrs]
+    elif not isinstance(arrs[0], np.ndarray):
         arrs = [np.asarray(ele) for ele in arrs]
+    else:
+        dtype = arrs[0].dtype if dtype is None else dtype
+
     original_length = [ele.shape[pad_axis] for ele in arrs]
     max_size = max(original_length)
+
     ret_shape = list(arrs[0].shape)
     ret_shape[pad_axis] = max_size
     ret_shape = (len(arrs), ) + tuple(ret_shape)
-    dtype = arrs[0].dtype if dtype is None else dtype
-    if use_shared_mem:
-        ret = mx.nd.full(shape=ret_shape, val=pad_val, ctx=mx.Context('cpu_shared', 0),
-                         dtype=dtype)
-        original_length = mx.nd.array(original_length, ctx=mx.Context('cpu_shared', 0),
-                                      dtype=np.int32)
-    else:
-        ret = mx.nd.full(shape=ret_shape, val=pad_val, dtype=dtype)
-        original_length = mx.nd.array(original_length, dtype=np.int32)
+
+    ret = np.full(shape=ret_shape, fill_value=pad_val, dtype=dtype)
+
     for i, arr in enumerate(arrs):
         if arr.shape[pad_axis] == max_size:
             ret[i] = arr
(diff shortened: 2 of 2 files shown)


leezu:
@szhengac could you comment on any effects on your nmt scripts? I'm not sure if you'll see a Speed-Up, given that the Batchify function runs in a separate process. However, at least CPU load should be significantly reduced.

leezu:
Initially I added a special case to use list comprehension if the input data is a list of lists. But there was no significant speedup compared to just converting the lists to numpy arrays. So I left that special case out to keep the code simple.

szhengac:
I am not sure about it. But I performed profiling before, and CPU time is negligible in NMT.

szhengac:
I think you can try sa script first to see how it affects the performance. sa will take much less time to run.

leezu:
Ok, sure. These are the results for sentiment_analysis script:

PR:
`python3 sentiment_analysis.py --gpu 0 --epochs 1  294.06s user 13.54s system 261% cpu 1:57.55 total`

Master branch:
`python3 sentiment_analysis.py --gpu 0 --epochs 1  399.97s user 20.40s system 302% cpu 2:19.13 total`

(Output of the [time](https://linux.die.net/man/1/time) command)

szhengac:
Can you measure the memory? because it will perform a few more copies.

leezu:
There won't be more methods. All usages of the Pad() function in GluonNLP currently have either Python list or numpy array as input. For these cases, we just delay the conversion to mxnet in this PR.
For the case where numpy array is passed as input (I believe in your script), the memory usage should actually be reduced as we don't convert every array to mxnet, but only the resulting padded array.

leezu:
These are the results including peak memory usage (obtained with `/usr/bin/time -v`).
Summary: Besides the CPU improvements noted above, and a reduction of voluntary context switches by half I don't see any significant differences for the sentiment analysis script.

Master branch:
 ```
        User time (seconds): 407.08
        System time (seconds): 21.05
        Percent of CPU this job got: 297%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 2:23.88
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 2651932
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 0
        Minor (reclaiming a frame) page faults: 860226
        Voluntary context switches: 542043
        Involuntary context switches: 208445
        Swaps: 0
        File system inputs: 0
        File system outputs: 57032
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
```

This PR:
```
        User time (seconds): 297.52
        System time (seconds): 13.18
        Percent of CPU this job got: 262%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 1:58.56
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 2653880
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 0
        Minor (reclaiming a frame) page faults: 753548
        Voluntary context switches: 212843
        Involuntary context switches: 208433
        Swaps: 0
        File system inputs: 0
        File system outputs: 57056
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0

```

**Comment by mli** [comment]
Job PR-249/6 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-249/6/index.html

**Comment by leezu** [comment]
Following this change, if we compose two padding functions, there is an un-needed conversion to mxnet and back to numpy. Consider:

```
pad0 = nlp.data.batchify.Pad(pad_val=subword_pad_val)
pad1 = nlp.data.batchify.Pad(axis=1, pad_val=subword_pad_val)
pad1([
  pad0([... for ... in x[i]])
  for i in range(...)
])
```

The output of pad0 will be an mxnet NDArray, which pad1 will then convert to numpy to efficiently pad the axis=1. The composition will still be significantly faster than before this PR, but we may still want to avoid it. However, I'm not sure how to avoid it without breaking the API. For a long-term vision, I hope we can fix the slowdown in mxnet and switch the Pad function back to use NDArray. We may choose to ignore above composition inefficiency until then?

szha:
This can happen outside the context of pad too so I think it's acceptable.

PR #268, opened 2018-08-12
fix weight drop and add tests

### LINKED ISSUE #264: apply_weight_drop required changes to integrate mxnet 1.3
our existing implementation on apply_weight_drop relied on a non-public attribute in gluon RNN blocks that were removed apache/incubator-mxnet#11482 when making the RNN blocks hybrid, and it does not have test, which will prevent us from integrating with mxnet 1.3. We need:
- [ ] fix apply_weight_drop and upgrade the mxnet dependency to allow 1.3 integration.
- [ ] add test for apply_weight_drop
- [ ] bump up minor version to 0.4 which relies on 1.3.

szha:
## Description ##
fix weight drop with new mxnet version

### Essentials ###
- [x] Code is well-documented

### Changes ###
- [x] update mxnet dependency to the latest nightly
- [x] fix weight drop
- [x] add weight drop tests

## Comments ##
- resolves #264

### DIFF
--- a/gluonnlp/model/utils.py
+++ b/gluonnlp/model/utils.py
@@ -30,7 +30,7 @@ from .parameter import WeightDropParameter
 # pylint: disable=too-many-nested-blocks
 def apply_weight_drop(block, local_param_regex, rate, axes=(),
                       weight_dropout_mode='training'):
-    """Apply weight drop to the parameter of a block.
+    r"""Apply weight drop to the parameter of a block.
 
     Parameters
     ----------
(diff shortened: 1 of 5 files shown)


tests/unittest/test_models.py
```diff
@@ -207,3 +206,68 @@ def test_big_rnn_model_share_params():
     logits, hidden = eval_model(x, hidden)
     assert logits.shape == (seq_len, batch_size, vocab_size)
     mx.nd.waitall()
+
+def test_weight_drop():
```
szhengac: Perhaps we also need to test whether the gradient is correct. We can do it by asserting the new parameter.

szha: added

mli:
Job PR-268/5 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-268/5/index.html

gluonnlp/model/utils.py
```diff
@@ -19,21 +19,25 @@
...
     """Apply weight drop to the parameter of a block.
 
     Parameters
     ----------
     block : Block or HybridBlock
         The block whose parameter is to be applied weight-drop.
-    local_param_name : str
-        The parameter name used on the block. such as 'weight'.
+    local_param_regex : str
+        The regex for parameter names used in the self.params.get(), such as 'weight'.
```
eric-haibin-lin: Can we update the doc with an example of using regex instead of exact match? such as ".*h2h_weight"

mli:
Job PR-268/10 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-268/10/index.html

PR #279, opened 2018-08-14
Add transformer notebook

cgraywang:
## Description ##
Add transformer notebook

### Essentials ###
- [ ] Code is well-documented

### Changes ###

## Comments ##


docs/examples/nmt/transformer.md
```diff
@@ -0,0 +1,603 @@
...
+scaled = True
+
+# parameters for training
+optimizer = 'adam'
+epochs = 3
+batch_size = 2700
+test_batch_size = 256
+num_accumulated = 1
+lr = 2
+warmup_steps = 1
```
szhengac: This can makes stepsize too large and possibly cause NAN.

cgraywang: It works fine for now, `step_size = float(loss_denom) / batch_size / 100.0`, since the `batch_size` is large. So you suggest use smaller `lr`?

szhengac:
If it works, then it is fine.

szhengac:
Why do we need to copy everything to example folder?

szha:
the notebook has a standalone download link on the webpage, in which the nmt script is not included.
the copy should be removed once encoder decoder api finishes and nmt models are moved into api.

docs/examples/nmt/nmt/dataset.py
```diff
@@ -0,0 +1,70 @@
...
+
+__all__ = ['TOY']
+
+import os
+from gluonnlp.data.translation import _TranslationDataset, _get_pair_key, _get_home_dir
+from gluonnlp.data.registry import register
+
+
+@register(segment=['train', 'val', 'test'])
+class TOY(_TranslationDataset):
```
szhengac: Don't use TOY dataset. This one is only created for testing the script. It is not calibrated with true training/testing sets.

szhengac: The performance in TOY dataset is not meaningful. This one only makes sure that the script is runnable. Instead, to present the meaningful result using the pre-trained model, you can consider  testing the first 100 sentences in testing set by setting a stop criterion in evaluation function.

cgraywang: Thanks for noticing this. But I have considered the case when creating this notebook, the notebook is just designed for showcase. The inference on WMT2014 will take more than 5mins, it is not possible to show to the audience in the tutorial. However, we have the dataset confirguration in the hyperparameter.py, we can tell the audience they can change the dataset from TOY to WMT2014 to reproduce the results. The notebook decription should have already reflected that. I have already synced this with Leo @leezu .

leezu: @cgraywang but according to @szhengac it seems that the Toy dataset is not meaningful, in fact it is not a "well known" dataset but just some test data that @szhengac extracted. It may not even follow the same preprocessing as WMT14. How about using the results on the first X sentences of WMT14 instead so that people don't get confused about the dataset part?

cgraywang: @szhengac @leezu  I just had a new commit which involves our discussion by using the WMT test dataset to show the results of the pretrained transformer. And use the TOY dataset just to showcase how to train the Transformer. Plus I also added the translation inference based on the user provided example and the Bleu score evaluation. Thanks.

szhengac: To allow the inference on user example, the model requires specific preprocessing pipeline on the sentence, including mosesdecoder processing shown in machine_translation.rst and applying BPE. As the learned BPE file is not stored, we need to preprocess the plain training set one more time and get the BPE file.

docs/examples/nmt/transformer.md
```diff
@@ -0,0 +1,698 @@
...
+  "text": "Translate the following English sentence into German:\n['We love each other']\nThe German translation is:\n['Wir lieben einander']\n"
+ }
+]
+```
+
+## Train Your Own Transformer
+
+In this subsection, we will go though the whole process about loading translation dataset in a more unified way, and create data sampler and loader, as well as define the Transformer model, finally writing training script to train the model yourself.
+
+### Load and Preprocess TOY Dataset
```
leezu: How about not using the TOY Dataset at all, because it is not a 'well-known' dataset but rather only exists inside GluonNLP for the purpose of speeding up some tests on the CI server. How about loading the WMT2014BPE here but in case that `demo = True`, throwing away all but the first X% of the dataset away?

docs/examples/nmt/nmt/dataset.py
cgraywang: @szhengac I have already written a simple inference function to do the data transformation and use beam search to generate the target sentence. Please check in the notebook example.

docs/examples/nmt/gnmt.md
```diff
@@ -297,20 +338,40 @@ loss_function = nmt.loss.SoftmaxCEMaskedLoss()
 loss_function.hybridize(static_alloc=static_alloc)
 ```
 
+```{.json .output n=7}
```
szha: don't include markdown's outputs. instead let it run

mli:
Job PR-279/30 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-279/30/index.html

scripts/tests/test_scripts.py
```diff
@@ -34,6 +36,7 @@ def test_embedding():
         '--epochs', '1', '--optimizer', 'sgd', '--ngram-buckets', '1000',
         '--max-vocab-size', '1000'
     ])
+    time.sleep(5)
```
eric-haibin-lin: `subprocess.check_call` already waits for the process to complete. Why is `time.sleep` required here?

cgraywang: The test_scripts might fail due to the multi-process testing. This might have a better chance to make the test run through.

docs/examples/nmt/transformer.md
cgraywang: Thanks for the suggestion @leezu . The reasons we are using TOY dataset instead of using WMT2014BPE are three folds: 1) The reason we are using TOY dataset is for showing how we can train the transformer in an unified data API. It is not trivial to use X% of the whole WMT2014BPE. We need to hack the data API, which will cause the API inconsistency compared to directly the data API usage in the   first portion of the notebook. 2) To hack the data API is not trivial: it requires getting the batch depends on the translation dataset transforms; similarly, creating the valid_lengths, samplers and data loader depends on the translation dataset. So we need to hack the translation dataset to use certain portion of the WMT dataset. Besides, we also need to hack the samplers and data loader. 3) Since as you know it is impossible to see the reasonable training result in the notebook. Despite the data loading issue, as you know it requires thousands of warm up steps. All in all, to show how to train the model, instead, simply using TOY dataset is enough and more reasonable.

leezu: Thanks for the explanation @cgraywang . I'm not sure if I can completely follow though. As far as I understand, WMT2014BPE is simply a gluon.data.ArrayDataset containing source and translated strings. So to get a subset of WMT2014BPE the following works:
```
>>> d = nlp.data.WMT2014BPE()
>>> first_thousand_src, first_thousand_target = d[:1000]
>>> small_d = mx.gluon.data.ArrayDataset(first_thousand_src, first_thousand_target)
>>> # or short mx.gluon.data.ArrayDataset(*d[:1000])
```

As far as I understand, we only need to delete below case from the `load_translation_data` function:
```
    elif dataset == 'TOY':
        common_prefix = 'TOY_{}_{}_{}_{}'.format(src_lang, tgt_lang,
                                                 hparams.src_max_len, hparams.tgt_max_len)
        data_train = nmt.dataset.TOY('train', src_lang=src_lang, tgt_lang=tgt_lang)
        data_val = nmt.dataset.TOY('val', src_lang=src_lang, tgt_lang=tgt_lang)
        data_test = nmt.dataset.TOY('test', src_lang=src_lang, tgt_lang=tgt_lang) 
```

and replace it with `mx.gluon.data.ArrayDataset(*d[:1000])` for train val and test.

To the interface cleaner, it may be nice to introduce some parameter, e.g. `k` to load_translation_data that allows to specify how many sentences at maximum to load (i.e. 1000 above) instead of simply returning only 1000 sentences if `demo=True`.


Given that this directly changes `load_translation_data`, I don't see why there would be any issues with valid lengths, sampler etc, as all of that should now be automatically be constructed with respect to the shorter dataset (ie. your points 2) and 3))

I believe this change would make the notebook easier to understand as there is no need for the TOY dataset here and it may confuse people reading the code.

cgraywang: Thanks @leezu for the quick response. I probably can make my explanation clearer... The purpose of the training script in the notebook is to show the users how to use a dataset (maybe user data, maybe not) to train their own transformer using the API already in the script. Then: 

1) if we are hacking the dataset the way you just showed, users will be confused, they probably will ask a) how to use my data set to train the model? will I need to split the dataset? do I need to change the code in the existing API (for example, `load_translation_data`)? This will break the existing API. 

2) The `TOY` dataset is essentially a sample from the WMT2014-EN-DE dataset. I show an example:

`data_train[0] = ('Orlando Bloom and Miranda Kerr still love each other', 'Orlando Bloom und Miranda Kerr lieben sich noch immer')`

3) If users turn on 'demo = False', it will take at least 5min to download the dataset and 1 day to make the training completed to get the SoTA results. If users intend to use the `demo` mode, the dataset should be small and show how to load the dataset without hacking the API. I guess we already agree on this point.

In conclusion, I believe use what we have right now is clear to the users.

szha: The download time is indeed a valid concern. For people with slower access to s3 (e.g., China), it might be a show stopper to force these users to sit there and wait for 30m just to get started.

On the other hand, our machine translation dataset already consists of different segments. For example:
https://github.com/dmlc/gluon-nlp/blob/master/gluonnlp/data/translation.py#L176-L192

It's conceivable that we can download one small segment for demonstration purpose. However, for the ease of implementation, we chose to package everything in a single file, which makes it impossible to download one single segment right now. I think the use-case in this PR is a reasonable justification to have per-segment download.

I will leave it to @cgraywang @leezu you two to decide how to proceed.

mli:
Job PR-279/35 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-279/35/index.html

docs/examples/nmt/transformer.md
leezu: I agree we should not download very large corpora for our example scripts. So for that reason I'm fine with using the toy corpus. I think it would be better to distribute large corpora in multiple files and have our API handle the case where only a subset of the dataset is requested. (But that is out of scope now). That would make the toy corpus obsolete.

Regarding point 1) and 2). I think on the contrary, coding up a new dataset class and doing the preprocessing manually (which is what is done for TOY), is not showing any strength of our dataset API. Instead it is nice to show how users can adapt the existing gluon-nlp library facilities to their needs.

szha: The item of breaking down the dataset is tracked in https://github.com/dmlc/gluon-nlp/issues/358

cgraywang: Thanks @leezu @szha for the helpful discussion on the notebook. I agree that if the large dataset can be downloaded in segments, it would be better and we can remove the TOY dataset entirely.

PR #307, opened 2018-08-19
Improve error message when loading pretrained fasttext with ngrams

leezu:
## Description ##

For some pretrained embeddings no subword / binary / ngram models were published
and trying to load them will fail. Currently "Cannot find pre-trained source X
for token embedding fasttext" will be raised, which is wrong given that the
source exists but simply no subword features are available for that source. Now
"No ngrams are available for {}." will be raised instead.

### Essentials ###
- [X] Code is well-documented

### Changes ###
- [X] Improve error message when loading pretrained fasttext with ngrams

## Comments ##
@eric-haibin-lin @szha we may want to do a patch release on Tuesday if we find more small usability issues like this

### DIFF
--- a/gluonnlp/embedding/token_embedding.py
+++ b/gluonnlp/embedding/token_embedding.py
@@ -925,7 +925,14 @@ class FastText(TokenEmbedding):
         self._check_source(self.source_file_hash, source)
 
         if load_ngrams:
-            self._check_source(self.source_bin_file_hash, source)
+            try:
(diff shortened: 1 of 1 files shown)


mli:
Job PR-307/2 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-307/2/index.html

PR #317, opened 2018-08-26
run tests in parallel

szha:
## Description ##
run tests in parallel

### Essentials ###
- [x] Code is well-documented

### Changes ###
- [x] run tests in parallel
- [x] mark stateful tests with serial mark
- [x] parameterize tests

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -64,12 +64,12 @@ stage("Unit Test") {
           prepare_clean_env('py2')
           install_dep('py2')
           code = 0
-          code += run_test('py2', 'tests/unittest', 'auto', 'not serial',
+          code += run_test('py2', 'tests/unittest', '4', 'not serial',
                    env.BRANCH_NAME.startsWith('PR-')?'':'--capture=no')
           code += run_test('py2', 'tests/unittest', '0', 'serial',
                    env.BRANCH_NAME.startsWith('PR-')?'':'--capture=no')
           report_cov(env.BRANCH_NAME+'-py2', 'unittests')
-          code += run_test('py2', 'scripts', 'auto', 'not serial',
+          code += run_test('py2', 'scripts', '4', 'not serial',
                    env.BRANCH_NAME.startsWith('PR-')?'':'--capture=no')
           code += run_test('py2', 'scripts', '0', 'serial',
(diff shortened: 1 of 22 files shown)


Jenkinsfile
```diff
@@ -28,7 +28,7 @@ stage("Unit Test") {
         python -m nltk.downloader all
         make clean
         python setup.py install --force
-        py.test -v --capture=no --durations=0 tests/unittest scripts
+        py.test -v -n auto --durations=0 tests/unittest scripts
```
leezu: I believe we need to call pytest two times now. Once with `-n auto -m "not serial"` and once with `-n0 -m "serial"`. See https://github.com/pytest-dev/pytest-xdist/issues/84#issuecomment-240608210.

By removing `--capture=no` the output of all passing tests will be removed. If you are confident we never need to look at the output of passing tests I'm happy to remove it. I added it as I wasn't sure if we may run into a situation where the output of previous successful runs would be helpful.

szha: OK. Will update the commands.

Currently, the test log is too verbose to navigate. There might be cases where previous logs would be helpful to narrow down a bug. How about we do `--capture=no` for master builds only?

leezu: It's ok to do `--capture=no` in general. Probably we can just regenerate the output for a particular test at a particular commit in case we need it.

Regarding the update, how about separating the "serial" and "parallel" stages into two Jenkins pipeline stages. In that way the "serial" and "parallel" pytest executions can run in parallel.

szha: We are running on only one host right now and `-n auto` would already make use of all CPUs, so we don't need to run these parts in parallel yet.

szha: Regarding the output capture, I don't think for each PR the current verbosity is needed, whereas for all master builds it might be useful to provide the output for debugging purpose, hence the suggestion for no capture on master only.

mli:
Job PR-317/14 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-317/14/index.html

mli:
Job PR-317/15 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-317/15/index.html

Issue #349, opened 2018-09-26
Transformer script doesn't run with py3


eric-haibin-lin:
```
MXNET_GPU_MEM_POOL_TYPE=Round python3 train_transformer.py --dataset WMT2014BPE                        -$
src_lang en --tgt_lang de --batch_size 2700                        --optimizer adam --num_accumulated 16 --lr 2.0 --warmup_steps 4000                        --save_dir transformer_$
n_de_u512 --epochs 30 --gpus 0,1,2,3,4,5,6,7 --scaled                        --average_start 5 --num_buckets 20 --bucket_scheme exp --bleu 13a --log_interval 10 2>&1 | tee base.log
```

```
2018-09-26 05:38:28,761 - root - Namespace(average_checkpoint=False, average_start=5, batch_size=2700, beam_size=4, bleu='13a', bucket_ratio=0.0, bucket_scheme='exp', dataset='WMT2$
14BPE', dropout=0.1, epochs=30, epsilon=0.1, full=False, gpus='0,1,2,3,4,5,6,7', hidden_size=2048, log_interval=10, lp_alpha=0.6, lp_k=5, lr=2.0, magnitude=3.0, num_accumulated=16,
num_averages=5, num_buckets=20, num_heads=8, num_layers=6, num_units=512, optimizer='adam', save_dir='transformer_en_de_u512', scaled=True, src_lang='en', src_max_len=-1, test_batc$
_size=256, tgt_lang='de', tgt_max_len=-1, warmup_steps=4000.0)
All Logs will be saved to transformer_en_de_u512/train_transformer.log
Load cached data from /home/ubuntu/transformer/scripts/nmt/cached/WMT2014BPE_en_de_-1_-1_train.npz
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/numpy/lib/format.py", line 650, in read_array
    array = pickle.load(fp, **pickle_kwargs)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xfa in position 8: ordinal not in range(128)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "train_transformer.py", line 304, in <module>
    = load_translation_data(dataset=args.dataset, src_lang=args.src_lang, tgt_lang=args.tgt_lang)
  File "train_transformer.py", line 260, in load_translation_data
    data_train_processed = load_cached_dataset(common_prefix + '_train')
  File "train_transformer.py", line 167, in load_cached_dataset
[...]

PR #350, opened 2018-09-27
Fix encoding for py2

szhengac:
## Description ##
Fix https://github.com/dmlc/gluon-nlp/issues/349

### Essentials ###
- [x] Code is well-documented

### Changes ###
- [x] Feature1, tests, (and when applicable, API doc)
- [x] Feature2, tests, (and when applicable, API doc)

### DIFF
--- a/scripts/nmt/train_transformer.py
+++ b/scripts/nmt/train_transformer.py
@@ -163,7 +163,7 @@ def load_cached_dataset(prefix):
     cached_file_path = os.path.join(_C.CACHE_PATH, prefix + '.npz')
     if os.path.exists(cached_file_path):
         print('Load cached data from {}'.format(cached_file_path))
-        dat = np.load(cached_file_path)
+        dat = np.load(cached_file_path, encoding='bytes')
         return ArrayDataset(np.array(dat['src_data']), np.array(dat['tgt_data']))
     else:
         return None
--- a/scripts/nmt/train_transformer.py
+++ b/scripts/nmt/train_transformer.py
@@ -163,7 +163,7 @@ def load_cached_dataset(prefix):
     cached_file_path = os.path.join(_C.CACHE_PATH, prefix + '.npz')
     if os.path.exists(cached_file_path):
         print('Load cached data from {}'.format(cached_file_path))
-        dat = np.load(cached_file_path, encoding='bytes')
+        dat = np.load(cached_file_path, encoding='latin1')
         return ArrayDataset(np.array(dat['src_data']), np.array(dat['tgt_data']))
     else:
         return None
--- a/scripts/nmt/bleu.py
+++ b/scripts/nmt/bleu.py
@@ -96,7 +96,7 @@ def _tokenize_mteval_13a(segment):
     norm = norm.replace('&lt;', '<')
     norm = norm.replace('&gt;', '>')
 
-    norm = ' {} '.format(norm)
+    norm = u' {} '.format(norm)
     norm = re.sub(r'([\{-\~\[-\` -\&\(-\+\:-\@\/])', ' \\1 ', norm)
     norm = re.sub(r'([^0-9])([\.,])', '\\1 \\2 ', norm)
     norm = re.sub(r'([\.,])([^0-9])', ' \\1 \\2', norm)
--- a/scripts/nmt/bleu.py
+++ b/scripts/nmt/bleu.py
@@ -204,8 +204,8 @@ def compute_bleu(reference_corpus_list, translation_corpus, tokenized=True,
             'references and translation should have format of list of list(list(str)) ' \
             'and list(list(str)), respectively, when toknized is True.'
     else:
-        assert isinstance(reference_corpus_list[0][0], str) and \
-               isinstance(translation_corpus[0], str), \
+        assert isinstance(reference_corpus_list[0][0], six.string_types) and \
+               isinstance(translation_corpus[0], six.string_types), \
             'references and translation should have format of list(list(str)) ' \
             'and list(str), respectively, when toknized is False.'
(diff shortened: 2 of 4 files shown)


scripts/nmt/train_transformer.py
```diff
@@ -163,7 +163,7 @@ def load_cached_dataset(prefix):
     cached_file_path = os.path.join(_C.CACHE_PATH, prefix + '.npz')
     if os.path.exists(cached_file_path):
         print('Load cached data from {}'.format(cached_file_path))
-        dat = np.load(cached_file_path)
+        dat = np.load(cached_file_path, encoding='bytes')
```
leezu: What if Py3 was used to save the file? Wouldn't encoding=bytes be wrong (ie. the correct encoding in that case would be utf-8)? 
I believe for other parts of the library we choose to default to 'utf-8' even if the user is running Py2. So instead of changing the load part maybe we should prevent writing something in bytes encoding?

szhengac: `np.load` does not support `encoding='utf-8'`, the default encoding is ascii.

szhengac: the cached dataset is stored in int32 format, so I think byte is ok

mli:
Job PR-350/4 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-350/4/index.html

scripts/nmt/train_transformer.py
leezu: What is stored in the numpy archive? Are these all strings? If so it may be better to save as np.array(X, dtype='O') and make sure that these are unicode strings in Py2. Otherwise all strings are padded to same length

szhengac: Arrays of index of word in vocabulary.

leezu: But why do you need to specify an encoding then? Encoding seems only useful when loading Python 2 generated pickled files in Python 3. Ie. it should not be necessary if your dtype is int32?

leezu:
Yes, if pickle is involved we need to make sure to pickle unicode strings when in Py2. But I currently don't follow where pickle is involved if you `np.load` call reads an array that only contains integers?

leezu:
Thanks for fixing this for now so that Haibin is unblocked!
I am still not sure though why it is necessary to specify an encoding for np.load, if 1) there is actually no string objects involved or 2) we rather can make sure to always use unicode.

szhengac:
np.load uses default encoding ascii to load the file. In this pr, I change
to latin1 which is a superset of ascii. In my testing, this has something
to do the with the stored np.array of np.array data structure. I have even
tried only fetch the first example of the stored dataset and save it again,
and then load it with ascii, but it still fail.

leezu:
Thanks @szhengac . But according to numpy API doc, the encoding parameter is "only useful when loading Python 2 generated pickled files in Python 3, which includes npy/npz files containing object arrays.". So if there is no need for pickle here, as you mentioned you only store 32bit integers, there should be no need for this option. Maybe there is something else going on?

szhengac:
I find that changing the encoding is necessary when the structure is
array([array(int), ..., array(int)) but not needed for array([[int], ...,
[int]]). perhaps the previous one requires some special symbol not in ascii.

leezu:
Yes, you are are storing an array of python objects (each object is a numpy array). You should use `np.savez` instead of `np.save` because the former allows you to save several arrays into a .npz archive. This will be much faster and will avoid these kind of Py2 / Py3 problems.

leezu:
`np.load` will happily read the npz archive file. You can find some examples at https://www.numpy.org/devdocs/reference/generated/numpy.savez.html https://www.numpy.org/devdocs/reference/generated/numpy.load.html

szhengac:
np.savez has been used

leezu:
You need to make sure that each argument to np.savez is a numpy array that does not contain any Python objects. I will take a look at your script why you have python objects in the arguments passed to savez.

leezu:
I suggest you replace

``` python
src_data = np.array([ele[0] for ele in dataset])
tgt_data = np.array([ele[1] for ele in dataset])
np.savez(os.path.join(_C.CACHE_PATH, prefix + '.npz'), src_data=src_data, tgt_data=tgt_data)
```
with
``` python
src_data = np.concatenate([e[0] for e in dataset])
tgt_data = np.concatenate([e[1] for e in dataset])
src_cumlen = np.cumsum([e[0] for e in dataset])
tgt_cumlen = np.cumsum([e[1] for e in dataset])
np.savez(os.path.join(_C.CACHE_PATH, prefix + '.npz'), src_data=src_data, tgt_data=tgt_data, src_cumlen=src_cumlen, tgt_cumlen=tgt_cumlen)
```

and the loading code respectively.

It will reduce the runtime of the code from ~80 seconds to ~4 seconds and avoid the pickle trouble.

leezu:
Also file-size will be reduced by 40%

leezu:
The numbers are for the train data that is used when running the script with default parameters.

szhengac:
We cannot use `np.concatenate` because the sequences have different lengths. What are for `src_cumlen` and `tgt_cumlen`.?

leezu:
concatenate concatenates a list of arrays. Consequently we have a single 1D array in your case. `src_cumlen` and `tgt_cumlen` store the cumulative lengths of the non-concatenated arrays. Based on this you can split the single, long 1D array up into the list of arrays of varying length when loading the cached data.

This allows you to efficiently serialize and deserialize the list of sequences of varying lengths. It only requires 5% of the time needed by the current implementation (because pickle is very expensive) and saves 40% space.

PR #359, opened 2018-10-12
Add GloVe

leezu:
## Description ##
This adds an implementation of [GloVe](https://nlp.stanford.edu/pubs/glove.pdf). Unlike the original C implementation with asynchronous AdaGrad we used synchronous batched AdaGrad to make use of GPU. This also adds optimized C++ tools `vocab_count` and `cooccur` to construct the input for the `train_glove.py` script. Both tools make use of a (single-machine) map-reduce pattern to scale better than the single-threaded versions of https://github.com/stanfordnlp/GloVe .

### Essentials ###
- [x] Code is well-documented

### Changes ###
- [X] Add GloVe
- [X] Scalable `vocab_count` and `cooccur` tools

## Comments ##
- The C++ tools require a compiler with basic C++17 support. They require a few few sub-modules for compilation: 1) [CNPY](https://github.com/leezu/cnpy) to write the output in numpy format. 2) [range-v3](https://github.com/ericniebler/range-v3/) Implementation of [Ranges-TS](https://en.cppreference.com/w/cpp/experimental/ranges) for C++ >= 11, 3) [sparsepp](https://github.com/greg7mdp/sparsepp) a [faster (less cache misses) and more memory efficient](https://github.com/greg7mdp/sparsepp/blob/master/bench.md#introducing-sparsepp) hashmap 4) [CLI11](https://github.com/CLIUtils/CLI11) Command line interface. CMake is set up to download all dependencies automatically when preparing the compilation of the tools.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,13 @@
+[submodule "scripts/word_embeddings/CLI11"]
+	path = scripts/word_embeddings/extern/CLI11
+	url = https://github.com/CLIUtils/CLI11.git
+[submodule "scripts/word_embeddings/cnpy"]
+	path = scripts/word_embeddings/extern/cnpy
+	url = https://github.com/leezu/cnpy
+	branch = libzip
+[submodule "scripts/word_embeddings/sparsepp"]
+	path = scripts/word_embeddings/extern/sparsepp
+	url = https://github.com/greg7mdp/sparsepp.git
+[submodule "scripts/word_embeddings/range-v3"]
+	path = scripts/word_embeddings/extern/range-v3
+	url = https://github.com/ericniebler/range-v3.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -8,6 +8,3 @@
 [submodule "scripts/word_embeddings/sparsepp"]
 	path = scripts/word_embeddings/extern/sparsepp
 	url = https://github.com/greg7mdp/sparsepp.git
-[submodule "scripts/word_embeddings/range-v3"]
-	path = scripts/word_embeddings/extern/range-v3
-	url = https://github.com/ericniebler/range-v3.git
--- a/scripts/word_embeddings/tools/cooccur.cc
+++ b/scripts/word_embeddings/tools/cooccur.cc
@@ -59,15 +59,16 @@ std::ostream &operator<<(std::ostream &in,
   return in << static_cast<int>(context_weight);
 }
 
+// Arguments specified via command line options. See ParseArgs for documentation.
 struct Arguments {
   unsigned int num_threads = 1;
-  unsigned int windowSize = 5;
-  bool noSymmetric = false;
+  unsigned int window_size = 5;
+  bool no_symmetric = false;
   bool subsample = false;
   ContextWeight context_weight;
 };
 
-auto parseArgs(int argc, char **argv) {
+auto ParseArgs(int argc, char **argv) {
   // Performance optimizations for writing to stdout
   std::ios::sync_with_stdio(false);
 
@@ -79,9 +80,9 @@ auto parseArgs(int argc, char **argv) {
   app.add_option("-o,--output", output,
                  "Output file name. Co-occurence matrix is saved as "
                  "scipy.sparse compatible CSR matrix in a numpy .npz archive");
-  app.add_option("-w,--window-size", args.windowSize,
+  app.add_option("-w,--window-size", args.window_size,
                  "Window size in which to count co-occurences.");
-  app.add_flag("--no-symmetric", args.noSymmetric,
+  app.add_flag("--no-symmetric", args.no_symmetric,
                "If not specified, a symmetric context window is used.");
   app.add_flag("--subsample", args.subsample,
                "Apply subsampling during co-occurence matrix construction as "
@@ -178,7 +179,7 @@ private:
 };
 
 // * Input
-auto readVocab() {
+auto ReadVocab() {
   std::string word;
   std::string count;
   int rank{0};
@@ -194,11 +195,11 @@ auto readVocab() {
 std::mutex paths_m;
 std::mutex matrices_m;
 
-void readMatrix(std::queue<fs::path> &paths, queue<Matrix> &matrices,
+void ReadMatrix(std::queue<fs::path> &paths, queue<Matrix> &matrices,
                 const Vocab &vocab, const Arguments &args, uint32_t seed) {
   assert(seed > 0);
   std::string line;
-  CircularBuffer<uint32_t> history(args.windowSize);
+  CircularBuffer<uint32_t> history(args.window_size);
   std::unique_ptr<Matrix> m = std::make_unique<Matrix>();
 
   // Prepare subsampling
@@ -255,7 +256,7 @@ void readMatrix(std::queue<fs::path> &paths, queue<Matrix> &matrices,
             uint64_t key; // We merge 32 bit row and col indices to a single 64
                           // bit key
             // For symmetric contexts, only store one direction.
-            if (!args.noSymmetric) {
+            if (!args.no_symmetric) {
               if (word_rank <= context_word_rank) {
                 key = (uint64_t)word_rank << 32 | context_word_rank;
               } else {
@@ -268,7 +269,7 @@ void readMatrix(std::queue<fs::path> &paths, queue<Matrix> &matrices,
             if (args.context_weight == ContextWeight::Harmonic) {
               (*m)[key] += 1.0f / static_cast<count_type>(distance);
             } else if (args.context_weight == ContextWeight::DistanceOverSize) {
-              (*m)[key] += (args.windowSize - distance - 1) / args.windowSize;
+              (*m)[key] += (args.window_size - distance - 1) / args.window_size;
             } else {
               (*m)[key]++;
             }
@@ -286,7 +287,7 @@ void readMatrix(std::queue<fs::path> &paths, queue<Matrix> &matrices,
   }
 }
 
-std::unique_ptr<Matrix> combineMatrices(queue<Matrix> &matrices,
+std::unique_ptr<Matrix> CombineMatrices(queue<Matrix> &matrices,
                                         int num_threads) {
   std::unique_ptr<Matrix> m1 = matrices.pop();
   for (int i = 1; i < num_threads; i++) {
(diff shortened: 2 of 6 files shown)


mli:
Job PR-359/3 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-359/3/index.html

mli:
Job PR-359/4 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-359/4/index.html

mli:
Job PR-359/5 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-359/5/index.html

mli:
Job PR-359/6 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-359/6/index.html

mli:
Job PR-359/7 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-359/7/index.html

scripts/word_embeddings/tools/cooccur.cc
```diff
@@ -0,0 +1,368 @@
...
+#include "cnpy.h"           // numpy
+#include "range/v3/all.hpp" // Ranges TS https://github.com/ericniebler/range-v3/
+#include "sparsepp/spp.h"   // fast sparse hash map
+
+#include "utils.h"
+
+namespace rsv = ranges::view;
+namespace fs = std::filesystem;
+using Vocab = spp::sparse_hash_map<std::string, std::pair<uint32_t, uint32_t>>;
+using count_type = float;
```
eric-haibin-lin: Is using float safe? float(50M) == float(50M+1) last time i checked with numpy

scripts/word_embeddings/tools/cooccur.cc
```diff
@@ -0,0 +1,368 @@
...
+  return in;
+}
+
+std::ostream &operator<<(std::ostream &in,
+                         const ContextWeight &context_weight) {
+  return in << static_cast<int>(context_weight);
+}
+
+struct Arguments {
+  unsigned int num_threads = 1;
```
eric-haibin-lin: It's strange to see the adoption of both camelCase and underscore_style.. Can we pick one? We can just follow https://google.github.io/styleguide/cppguide.html

scripts/word_embeddings/tools/vocab_count.cc
```diff
@@ -0,0 +1,151 @@
...
+  for (auto &file : files) {
+    paths.emplace(file);
+  }
+  std::vector<std::thread> threads;
+  queue<Vocab> vocabs;
+  for (unsigned int i = 0; i < numThreads; i++) {
+    threads.push_back(std::thread(
+        [&paths, &vocabs]() { readVocab(std::ref(paths), std::ref(vocabs)); }));
+  }
+  std::unique_ptr<Vocab> vocab = combineVocabs(vocabs, numThreads);
```
eric-haibin-lin: It is required that `combineVocabs` starts after all `readVocab` threads are done? Is that enforced by the blocking queue?

scripts/word_embeddings/tools/utils.h
```diff
@@ -0,0 +1,31 @@
+#pragma once
+
+#include <condition_variable>
+#include <deque>
+#include <memory>
+#include <mutex>
+#include <utility>
+
+// Based on https://stackoverflow.com/a/12805690
+template <typename T> class queue {
```
eric-haibin-lin: Add some documentation regarding the behavior for push and pop

scripts/word_embeddings/tools/cooccur.cc
```diff
@@ -0,0 +1,368 @@
...
+  unsigned int windowSize = 5;
```
eric-haibin-lin: Could you add some doc for what windowSize means?

scripts/word_embeddings/tools/cooccur.cc
leezu: In this case we don't really care if the count is 1 or 2 different. Given that the counts are based on a noisy corpus. So the memory saving of float vs double seems more important.

.gitmodules
```diff
@@ -0,0 +1,13 @@
+[submodule "scripts/word_embeddings/CLI11"]
+	path = scripts/word_embeddings/extern/CLI11
+	url = https://github.com/CLIUtils/CLI11.git
+[submodule "scripts/word_embeddings/cnpy"]
+	path = scripts/word_embeddings/extern/cnpy
+	url = https://github.com/leezu/cnpy
```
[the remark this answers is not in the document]
leezu: There is a bug in the original project. It does not support Zip64 standard, meaning that if the numpy archive is > 4GB it will write a invalid file.. I just replace the low-level zlip with the high-level libzip in https://github.com/rogersce/cnpy/pull/41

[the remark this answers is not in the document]
leezu: Ie. libzip handles all these Zip-standard details for us.

scripts/word_embeddings/tools/vocab_count.cc
leezu: Yes, enforced by blocking Queue. This is better than starting combineVocabs after all readVocabs are done, because the combination can run in parallel with the reading.

scripts/word_embeddings/tools/cooccur.cc
leezu: These are documented in the help strings declared in the parseArgs function a few lines below. I'll add a comment to point this out. Thanks

leezu:
Will update this PR (rebase) to follow the structure adopted in https://github.com/dmlc/gluon-nlp/pull/384 once merged.

mli:
Job PR-359/8 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-359/8/index.html

scripts/word_embeddings/tools/cooccur.cc
```diff
@@ -0,0 +1,369 @@
...
+    double t = 1E-4;
+    for (const double count : vocab | rsv::values | rsv::values) {
+      idx_to_pdiscard.push_back(1 - std::sqrt(t / (count / sum_counts)));
+    }
+  }
+
+  while (true) {
+    fs::path path;
+    {
+      std::scoped_lock lock(paths_m);
```
eric-haibin-lin: std::scoped_lock is cpp17. Old compilers don't support this. What about just using std::lock_guard?

scripts/word_embeddings/tools/cooccur.cc
```diff
@@ -0,0 +1,369 @@
...
+
+    std::ifstream in{path};
+    if (!in.is_open()) {
+      throw std::invalid_argument(path.string() + " cannot be opened!");
+    }
+    while (std::getline(in, line)) {
+      history.clear(); // Discard context from other lines
+
+      auto words = line | rsv::split((int (*)(int)) & std::isspace);
+      for (auto word : words) {
```
eric-haibin-lin: also, use `(const string & word : words)` to reduce a copy (and make it read only)

scripts/word_embeddings/tools/cooccur.cc
```diff
@@ -0,0 +1,369 @@
...
+
+          for (const auto &[distance, context_word_rank] :
+               rsv::zip(rsv::iota(1, static_cast<int>(history.size()) + 1),
+                        history)) {
+            uint64_t key; // We merge 32 bit row and col indices to a single 64
+                          // bit key
+            // For symmetric contexts, only store one direction.
+            if (!args.no_symmetric) {
+              if (word_rank <= context_word_rank) {
+                key = (uint64_t)word_rank << 32 | context_word_rank;
```
eric-haibin-lin: https://google.github.io/styleguide/cppguide.html#Casting cast with `static_cast<>`

scripts/word_embeddings/tools/cooccur.cc
```diff
@@ -0,0 +1,369 @@
...
+    }
+
+    std::ifstream in{path};
+    if (!in.is_open()) {
+      throw std::invalid_argument(path.string() + " cannot be opened!");
+    }
+    while (std::getline(in, line)) {
+      history.clear(); // Discard context from other lines
+
+      auto words = line | rsv::split((int (*)(int)) & std::isspace);
```
eric-haibin-lin: What does the `|` operator do for string?

eric-haibin-lin: I also don't quite understand what `(int (*)(int)) & std::isspace` means..

scripts/word_embeddings/tools/cooccur.cc
leezu: I'm not sure if this helps, in C++11 and later `std::string` is guaranteed to end with a `\0`, so by using `std::string` for every token here, we always force a copy of the tokens. C++17 introduces std::string_view as a workaround, but until https://wg21.link/P0919 is merged for upcoming C++20 we can't use `std::string_view` as a key in unordered containers that maps from `std::string` to X.
That's my understanding, but please let me know if you know anything else

scripts/word_embeddings/tools/cooccur.cc
leezu: `|` here is part of [Ranges TS](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1037r0.pdf). It will be part of C++20 according to the Standards Meeting in Rapperswil this summer (included here via `#include "range/v3/all.hpp"` which runs on C++11)

leezu: `line | rsv::split((int (*)(int)) & std::isspace);` returns a range over all tokens in the line.

> Given a source range and a delimiter specifier, split the source range into a range of ranges using the delimiter specifier to find the boundaries. The delimiter specifier can be a value, a subrange, a predicate, or a function. The predicate should take an single argument of the range's reference type and return true if and only if the element is part of a delimiter. The function should accept current/end iterators into the source range and return make_pair(true, iterator_past_the_delimiter) if the current position is a boundary; otherwise, make_pair(false, cur). The delimiter character(s) are excluded from the resulting range of ranges

Unfortunately, the lack of https://wg21.link/P0919 means we need to copy each range (token) to a `std::string` in the end, but once https://wg21.link/P0919 is supported, using range with string_view would avoid any copies here.

Some more example code for splitting ranges is at https://github.com/ericniebler/range-v3/blob/f662cbc3f8643cc4d8daaab0b53b9b1daddf663b/test/view/split.cpp#L146-L157

scripts/word_embeddings/tools/cooccur.cc
eric-haibin-lin: Does it still create a copy with a reference `&`?

mli:
Job PR-359/9 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-359/9/index.html

mli:
Job PR-359/10 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-359/10/index.html

mli:
Job PR-359/12 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-359/12/index.html

PR #369, opened 2018-10-20
ConcatDataset

hhexiy:
## Description ##
Concatenate multiple datasets; datasets can be of different classes.

### Essentials ###
- [x] Code is well-documented

### Changes ###
- [x] ConcatDataset

### DIFF
--- a/gluonnlp/data/dataset.py
+++ b/gluonnlp/data/dataset.py
@@ -20,14 +20,40 @@
 # pylint: disable=undefined-all-variable
 """NLP Toolkit Dataset API. It allows easy and customizable loading of corpora and dataset files.
 Files can be loaded into formats that are immediately ready for training and evaluation."""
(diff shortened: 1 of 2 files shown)


gluonnlp/data/dataset.py
```diff
@@ -20,16 +20,42 @@
...
+    """Dataset that concatenates a list of datasets.
+
+    Parameters
+    ----------
+    datasets : list
+        List of datasets.
+    """
+    def __init__(self, datasets):
+        self.datasets = datasets
+        self.cum_sizes = np.cumsum([len(d) for d in datasets])
```
szha:
```suggestion
        self.cum_sizes = np.cumsum([0] + [len(d) for d in datasets])
```

leezu:
Yes, it is not required, however we may still support it as long as there is no downside

szha:
#381

PR #380, opened 2018-10-26
add sentencepiece tokenizer detokenizer

szha:
## Description ##
add sentencepiece tokenizer detokenizer

### Essentials ###
- [x] Code is well-documented

### Changes ###
- [x] Add sentencepiece tokenizer and detokenizer to support loading pre-trained subword models from google/sentencepiece

### DIFF
--- a/env/doc.yml
+++ b/env/doc.yml
@@ -19,3 +19,4 @@ dependencies:
     - mxnet-cu90>=1.3.0
     - sacremoses
     - seaborn
+    - sentencepiece
--- a/env/py2.yml
+++ b/env/py2.yml
@@ -14,3 +14,4 @@ dependencies:
   - pytest-xdist
   - pip:
     - mxnet-cu90>=1.3.0
+    - sentencepiece
--- a/env/py3.yml
+++ b/env/py3.yml
@@ -15,3 +15,4 @@ dependencies:
   - pip:
     - mxnet-cu90>=1.3.0
     - sacremoses
+    - sentencepiece
--- a/setup.py
+++ b/setup.py
@@ -60,6 +60,7 @@ setup(
             'scipy',
             'numba>=0.40.1',
             'jieba',
+            'sentencepiece',
         ],
         'dev': [
             'pytest',
--- a/src/gluonnlp/data/transforms.py
+++ b/src/gluonnlp/data/transforms.py
@@ -25,7 +25,7 @@ from __future__ import print_function
 
 __all__ = ['ClipSequence', 'PadSequence', 'SacreMosesTokenizer', 'NLTKMosesTokenizer',
            'SpacyTokenizer', 'SacreMosesDetokenizer', 'NLTKMosesDetokenizer', 'JiebaTokenizer',
-           'NLTKStanfordSegmenter']
+           'NLTKStanfordSegmenter', 'SentencepieceTokenizer', 'SentencepieceDetokenizer']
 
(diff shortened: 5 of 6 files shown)


mli:
Job PR-380/2 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-380/2/index.html

gluonnlp/data/transforms.py
```diff
@@ -668,3 +668,131 @@ def __call__(self, sample):
...
+    path : str
+        Path to the pre-trained subword tokenization model.
+
+    Examples
+    --------
+    >>> tokenizer = SentencepieceTokenizer('./corpus.full.bpe')
+    >>> detokenizer = SentencepieceDetokenizer('./corpus.full.bpe')
+    >>> sentence = 'This is a very awesome, life-changing sentence.'
+    >>> tokenizer(sentence)
+    ['▁This',
```
szhengac: This looks different from the moses BPE, in which subwords are identified by symbols @@. For example, with moses BPE, '_This' will be 'This' and 'ch' will be 'ch@@'.

szha: Yes indeed, and I wasn't able to find an option to set the word boundary markers. Is there such option in moses bpe?

szha: On the other hand, we may choose to integrate https://github.com/rsennrich/subword-nmt to support both, which carries an MIT license.

mli:
Job PR-380/4 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-380/4/index.html

gluonnlp/data/transforms.py
szha: I will address this in a separate PR.

leezu:
Thanks @szha. The integration works great.

For the future, I wonder: SentencePiece manages it's own Piece - ID mapping, which we may wan't to use as well. Would it make sense to - in addition to this - introduce a SentencePiece wrapper for our Vocab class? My concern here is that based on a particular tokenization obtained from the tokenizer in this PR, there is no guarantee that all possible tokens the sentencepiece model knows about are contained. So constructing a vocabulary on the gluonnlp side may not make sense. By treating sentencepiece as a vocabulary implementation it seems we may also expose `SampleEncodeAsIds` etc.

szha:
@leezu I think it makes sense to integrate the vocabulary of sentencepiece, but maybe not in the way as a wrapper around the specific functionality. On one hand, the serialized format of vocab in sentencepiece is a simple piece-ID mapping, which can already be covered by the vocab functionality in gluonnlp. On the other hand, if we choose to extend the functionality of our own vocab, we may run into difficulties to make such wrapper compatible with extended interface. Thus, I'd prefer to have the integration to be at the level of serialization mapping rather than wrapper class.

szha:
Actually there's a simple way to get all the tokens without a vocab file.

mli:
Job PR-380/8 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-380/8/index.html

mli:
Job PR-380/10 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-380/10/index.html

PR #397, opened 2018-11-11
Fix LazyTransformDataStream prefetching

leezu:
## Description ##
For some `DataStream` `s1`, `s2 = s.transform(fn)` will lazily apply `fn` to all elements of `s1` when `s2` is iterated. For this purpose `s.transform(fn)` returns a `_LazyTransformDataStream` object which implements `__iter__` to achieve the behaviour described above.

However, currently _LazyTransformDataStream `__iter__` is implemented with a generator (`yield`). Consequently, when calling `i = iter(s2)`, `iter(s1)` is not called. `iter(s1)` is only called upon `next(i)`. This is bad, as for if s1 is a `PrefetchingStream`, it will only start prefetching once `iter(s1)` is called.

### Essentials ###
- [X] Code is well-documented

### Changes ###
- [X] Fix prefetching behaviour when lazily transforming a `PrefetchingStream`

## Comments ##
- We should establish a "formal" requirement for all `DataStream`, to call `__iter__` on all wrapped streams if their `__iter__` method is called.
- Will add tests soon

### DIFF
--- a/src/gluonnlp/data/stream.py
+++ b/src/gluonnlp/data/stream.py
@@ -103,19 +103,27 @@ class _LazyTransformDataStream(DataStream):
 
     def __iter__(self):
         stream_iter = iter(self._stream)
-        try:
-            item = next(stream_iter)
-        except StopIteration:
-            return
-        istuple = isinstance(item, tuple)
-        if istuple:
-            yield self._fn(*item)
-            for item in stream_iter:
+
+        # Yield must be hidden in closure so that __iter__ is called before
+        # __next__ is called. This is important, as calling iter(self._stream)
+        # may trigger multi-threaded or multi-processing prefetching of the
+        # stream.
+        def _closure():
+            try:
+                item = next(stream_iter)
+            except StopIteration:
+                return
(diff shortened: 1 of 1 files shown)


mli:
Job PR-397/1 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-397/1/index.html

eric-haibin-lin:
why would a closure fix this?

leezu:
Because according to [PEP 255](https://www.python.org/dev/peps/pep-0255/)

> When a generator function is called, the actual arguments are bound to function-local formal argument names in the usual way, but no code in the body of the function is executed. Instead a generator-iterator object is returned; this conforms to the iterator protocol [6], so in particular can be used in for-loops in a natural way. Note that when the intent is clear from context, the unqualified name "generator" may be used to refer either to a generator-function or a generator-iterator.

Currently `__iter__` is a generator function. With the closure, it remains a normal function and is executed, returning the executed closure after triggering `iter(self._stream)`.

PR #406, opened 2018-11-12
Propagate exceptions from PrefetchingStream

leezu:
## Description ##
This adds logic to PrefetchingStream to report exceptions during setup and prefetching. Based on that, tests for #397 are added. 

### Essentials ###
- [X] Code is well-documented

### Changes ###
- [X] Propagate exceptions from PrefetchingStream workers

## Comments ##
- While checking for any other DataStreams that may suffer from similar issues as in #397 I found an unused file in the scripts folder and deleted it.
- Also wraps the generator  in `_EmbeddingCenterContextBatchify.__iter__`. While an implementation detail (unlike the lazy transformation of streams), it improves consistency throughout the codebase


src/gluonnlp/data/stream.py
```diff
@@ -221,7 +221,11 @@ def run(self):
         """Method representing the process’s activity."""
         random.seed(self.seed)
         np.random.seed(self.np_seed)
-        mx.random.seed(self.mx_seed)
+        if not isinstance(self, multiprocessing.Process):
+            # Calling mxnet methods in a subprocess will raise an exception if
+            # mxnet is built with GPU support
+            # https://github.com/apache/incubator-mxnet/issues/4659
+            mx.random.seed(self.mx_seed)
```
leezu: @szha @eric-haibin-lin FYI, we cannot guarantee the execution of any mxnet related code would work inside a subprocess.

mli:
Job PR-406/9 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-406/9/index.html

src/gluonnlp/data/stream.py
```diff
@@ -213,57 +221,97 @@ def run(self):
...
+            return next_item
+        else:
+            self._controlq.put(None)
+            if isinstance(next_error[0], StopIteration):
+                raise StopIteration
+            else:
+                return self._reraise(*next_error)
+
+    def _reraise(self, e, tb):
+        print('Reraising exception from Prefetcher')
```
zhreshold: is print appropriate here?

src/gluonnlp/data/stream.py
```diff
@@ -213,57 +221,97 @@ def run(self):
...
+        print(tb)
+        raise e
+
+    def _check_start(self):
+        next_error = self._errorq.get(block=True)
```
zhreshold: if _check_start is called multiple times accidentally, it's going to cause deadlock?

src/gluonnlp/data/stream.py
leezu: The idea is to replicate the standard behaviour of Python printing a traceback before crashing due to an exception. So I guess `print` is appropriate, but let me know if you have another idea. We should `print` to `stderr`. I will push a change for that, thanks for catching this

src/gluonnlp/data/stream.py
leezu: When calling it twice the behaviour is undefined and may cause deadlock (when reaching the end of the stream). I will add an assertion that it is not called twice. Thanks!

leezu:
Rebased, force-pushed and added a commit to address @zhreshold comments. @zhreshold , once you're happy with the code, can you approve the PR so it can be merged? If you have any other comments, please let me know

zhreshold:
@leezu LGTM now. Thanks!

PR #407, opened 2018-11-12
Fix tie_weight in nmt

szhengac:
## Description ##
Change self.tgt_embed.params to self.tgt_embed[0].params

### Essentials ###
- [x] Code is well-documented

## Comments ##


szha:
Also, does it mean we need to re-train the model?

szhengac:
What kind of test do you think is appropriate? Yes, we need to retrain the model. I think we need to postpone this pr until we get a new model, otherwise the loading of pretrained model can cause error.

szha:
@szhengac I'm merging this now as this bug fix is needed for another PR. Please still open another PR and add test.

PR #412, opened 2018-11-16
Add BiDAF question answering model

Ishitori:
## Description ##
Add Bidirectional Attention Flow model from [the original paper](https://arxiv.org/pdf/1611.01603.pdf). Current implementation achieves similar results with hyperparameters taken from the paper (for exact values look `get_args()` method in `train_question_answering.py`) on a single gpu: EM=67.1 and F1=76.9. This implementation also supports multigpu.

Default arguments of the code match to hyperparameters of the paper. One can run the training script in 3 different modes: data preprocessing, training, evaluation (last two can be combined). If data preprocessing step hasn't been done before running training, the data would be automatically preprocessed. For common usage, the following command line is enough (run from gluon-nlp root folder):

```
python3 -m scripts.question_answering.train_question_answering --train --evaluate --preprocessed_dataset_path scripts/question_answering/preprocessed_dataset.p --preprocessed_val_dataset_path scripts/question_answering/preprocessed_dataset_val.p --save_prediction_path scripts/question_answering/output/predictions.txt  --save_dir scripts/question_answering/output/ --word_vocab_path scripts/question_answering/squad_word_vocab.p --char_vocab_path scripts/question_answering/squad_char_vocab.p --gpu 0
```

Majority of code review was done in a [separate fork](https://github.com/cgraywang/gluon-nlp-1/pull/15). Some action items are planned to be done before merging with master branch (see "Next step" in the original review)

### Essentials ###
- [X] Code is well-documented

### Changes ###
- [X] Add BiDAF model and a training + evaluation scripts

### DIFF
--- a/scripts/question_answering/attention_flow.py
+++ b/scripts/question_answering/attention_flow.py
@@ -1,0 +1,70 @@
+# coding: utf-8
+
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+"""Attention Flow Layer"""
+from mxnet import gluon
+
+from .similarity_function import DotProductSimilarity
+
+
+class AttentionFlow(gluon.HybridBlock):
+    """
+    This ``block`` takes two ndarrays as input and returns a ndarray of attentions.
+
+    We compute the similarity between each row in each matrix and return unnormalized similarity
+    scores.  Because these scores are unnormalized, we don't take a mask as input; it's up to the
+    caller to deal with masking properly when this output is used.
+
+    By default similarity is computed with a dot product, but you can alternatively use a
+    parameterized similarity function if you wish.
+
+
+    Input:
+        - ndarray_1: ``(batch_size, num_rows_1, embedding_dim)``
+        - ndarray_2: ``(batch_size, num_rows_2, embedding_dim)``
+
+    Output:
+        - ``(batch_size, num_rows_1, num_rows_2)``
+
+    Parameters
+    ----------
+    similarity_function: ``SimilarityFunction``, optional (default=``DotProductSimilarity``)
+        The similarity function to use when computing the attention.
+    """
+    def __init__(self, similarity_function, batch_size, passage_length,
+                 question_length, embedding_size, **kwargs):
+        super(AttentionFlow, self).__init__(**kwargs)
+
+        self._similarity_function = similarity_function or DotProductSimilarity()
+        self._batch_size = batch_size
+        self._passage_length = passage_length
+        self._question_length = question_length
+        self._embedding_size = embedding_size
+
+    def hybrid_forward(self, F, matrix_1, matrix_2):
+        # pylint: disable=arguments-differ
+        tiled_matrix_1 = matrix_1.expand_dims(2).broadcast_to(shape=(self._batch_size,
+                                                                     self._passage_length,
+                                                                     self._question_length,
+                                                                     self._embedding_size))
+        tiled_matrix_2 = matrix_2.expand_dims(1).broadcast_to(shape=(self._batch_size,
+                                                                     self._passage_length,
+                                                                     self._question_length,
+                                                                     self._embedding_size))
+        return self._similarity_function(tiled_matrix_1, tiled_matrix_2)
--- a/scripts/question_answering/attention_flow.py
+++ b/scripts/question_answering/attention_flow.py
@@ -58,7 +58,7 @@ class AttentionFlow(gluon.HybridBlock):
         self._embedding_size = embedding_size
 
     def hybrid_forward(self, F, matrix_1, matrix_2):
-        # pylint: disable=arguments-differ
+        # pylint: disable=arguments-differ,unused-argument,missing-docstring
         tiled_matrix_1 = matrix_1.expand_dims(2).broadcast_to(shape=(self._batch_size,
                                                                      self._passage_length,
                                                                      self._question_length,
--- a/scripts/question_answering/attention_flow.py
+++ b/scripts/question_answering/attention_flow.py
@@ -47,24 +47,20 @@ class AttentionFlow(gluon.HybridBlock):
     similarity_function: ``SimilarityFunction``, optional (default=``DotProductSimilarity``)
         The similarity function to use when computing the attention.
     """
-    def __init__(self, similarity_function, batch_size, passage_length,
-                 question_length, embedding_size, **kwargs):
+    def __init__(self, similarity_function, passage_length,
+                 question_length, **kwargs):
         super(AttentionFlow, self).__init__(**kwargs)
 
         self._similarity_function = similarity_function or DotProductSimilarity()
-        self._batch_size = batch_size
         self._passage_length = passage_length
         self._question_length = question_length
-        self._embedding_size = embedding_size
 
     def hybrid_forward(self, F, matrix_1, matrix_2):
         # pylint: disable=arguments-differ,unused-argument,missing-docstring
-        tiled_matrix_1 = matrix_1.expand_dims(2).broadcast_to(shape=(self._batch_size,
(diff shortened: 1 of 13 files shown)


mli:
Job PR-412/6 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-412/6/index.html

scripts/question_answering/question_answering.py
```diff
@@ -0,0 +1,324 @@
...
+
+        # Pass through highway, shape remains unchanged
+        highway_output, _ = F.contrib.foreach(highway, highway_input, [])
+
+        # Transpose to TNC - default for LSTM
+        ce_output = self._contextual_embedding(highway_output)
+        return ce_output
+
+
+class BiDAFModelingLayer(HybridBlock):
```
szha: This class doesn't seem to add much value.

scripts/question_answering/question_id_mapper.py
```diff
@@ -0,0 +1,63 @@
...
+"""Question id mapper to and from int"""
+
+
+class QuestionIdMapper:
+    """Stores mapping between question id and context of SQuAD dataset
+
+    """
+    def __init__(self, dataset):
+        self._question_id_to_context = {item[1]: item[3] for item in dataset}
+        self._question_id_to_idx = {item[1]: item[0] for item in dataset}
```
szha: This assumes a schema that wasn't documented.

scripts/question_answering/similarity_function.py
```diff
@@ -0,0 +1,198 @@
...
+    def hybrid_forward(self, F, array_1, array_2):
+        result = (array_1 * array_2).sum(axis=-1)
+
+        if self._scale_output:
+            result *= F.sqrt(array_1.shape[-1])
+
+        return result
+
+
+class CosineSimilarity(SimilarityFunction):
```
szha: This along with several others are available in https://github.com/dmlc/gluon-nlp/blob/master/src/gluonnlp/embedding/evaluation.py#L29

scripts/question_answering/question_answering.py
Ishitori: Yes, but I would prefer to leave it as there is such an entity in the paper: picture and a reference block of text. So, for the ease of understanding, I wouldn't remove it.

Ishitori:
@szha, the paper reports numbers for dev and test set. Since test set is not publicly available, I was aiming to reach dev set numbers, which are EM=67.7, F1=77.3 (see table 1b, not 1a). 

My numbers are still lower a bit than these ones. I wouldn't be obsessed too much with that though. From my experience, these numbers can fluctuate a bit even in official implementation: when I was running original code in TF, using provided terminal commands, I got only {"f1": 76.64064976048196, "exact_match": 67.19016083254493}, which are lower than reported in the paper.

scripts/question_answering/question_id_mapper.py
Ishitori: Well, it says above, that it uses SQuAD dataset as an argument. The schema of SQuAD dataset is described in `gluonnlp/data/question_answering.py`. 

We certainly can improve it by providing more comments, if needed.

scripts/question_answering/similarity_function.py
Ishitori: Yes, looks like a duplicate, and in this case, other similarity functions should be put in the same place as the rest of them. 

Let's do it once we will move BiDAF and QANet into the master branch.

scripts/question_answering/question_answering.py
szha: You can do that in the BiDAFModel class since `self.modeling_layer = gluon.rnn.LSTM(...)` is as readable as `self.modeling_layer = BiDAFModelingLayer()`

scripts/question_answering/attention_flow.py
```diff
@@ -0,0 +1,70 @@
...
+
+    Output:
+        - ``(batch_size, num_rows_1, num_rows_2)``
+
+    Parameters
+    ----------
+    similarity_function: ``SimilarityFunction``, optional (default=``DotProductSimilarity``)
+        The similarity function to use when computing the attention.
+    """
+    def __init__(self, similarity_function, batch_size, passage_length,
```
szha: gluon blocks don't usually take batch sizes. try removing it from all classes.

scripts/question_answering/bidaf.py
```diff
@@ -0,0 +1,121 @@
...
+        return np.finfo(np.float32).min
+
+    def _get_small_positive_value(self):
+        """Provides minimal possible Float32 value
+        Returns
+        -------
+        value : float32
+            Minimal float32 value
+        """
+        return np.finfo(np.float32).eps
```
leezu: Would it make sense to use `np.finfo(dtype).eps`. The model may be trained with `float16`?

scripts/question_answering/question_answering.py
```diff
@@ -0,0 +1,324 @@
...
+            self._end_index_lstm = LSTM(hidden_size=span_start_input_dim,
+                                        num_layers=nlayers, dropout=dropout, bidirectional=biflag,
+                                        input_size=2 * span_start_input_dim)
+            self._end_index_combined = nn.Dense(units=1, in_units=8 * span_start_input_dim,
+                                                flatten=False)
+            self._end_index_model = nn.Dense(units=1, in_units=2 * span_start_input_dim,
+                                             flatten=False)
+
+    def hybrid_forward(self, F, x, m, mask, *args):
+        # pylint: disable=arguments-differ,unused-argument,missing-docstring
```
leezu: Why `unused-argument` must be disabled?

leezu: Same question for the other uses in this script

scripts/question_answering/train_question_answering.py
```diff
@@ -0,0 +1,769 @@
...
+        Learning rate
+    """
+    if options.resume_training:
+        return options.lr
+
+    return min(options.lr, options.lr * (math.log(iteration) / math.log(options.lr_warmup_steps)))
+
+
+def get_gradients(model, ctx, options):
+    """Get gradients and apply gradient decay to all layers if required.
```
leezu: Is this different to the weight decay provided by the mxnet optimizer?

scripts/question_answering/train_question_answering.py
```diff
@@ -0,0 +1,769 @@
...
+
+    net.collect_params().load(params_path, ctx=ctx)
+    net.hybridize(static_alloc=True)
+    return evaluator.evaluate_performance(net, ctx, options)
+
+
+def get_args():
+    """Get console arguments
+    """
+    parser = argparse.ArgumentParser(description='Question Answering example using BiDAF & SQuAD')
```
leezu: Please add `formatter_class=argparse.ArgumentDefaultsHelpFormatter`

scripts/question_answering/utils.py
```diff
@@ -0,0 +1,379 @@
...
+        elif operation == '/':
+            return first_tensor / second_tensor
+        elif operation == '+':
+            return first_tensor + second_tensor
+        elif operation == '-':
+            return first_tensor - second_tensor
+        else:
+            raise NotImplementedError
+
+
```
leezu: Would it make sense to make the functions below `HybridBlocks`?

scripts/question_answering/question_answering.py
Ishitori: Removed the class

scripts/question_answering/question_id_mapper.py
Ishitori: Added pydoc to the __init__ method that dataset is a SQuAD dataset

scripts/question_answering/attention_flow.py
Ishitori: Removed from every block

scripts/question_answering/bidaf.py
Ishitori: It was supported, but we actually never tested the convergence and remove all code for fp16. I left it here because it still provides meaningful name.

scripts/question_answering/question_answering.py
Ishitori: I think it was because of *args in `hybrid_forward`. Removed it everywhere

scripts/question_answering/train_question_answering.py
Ishitori: A little bit. This particular method applies it only to UNK token of the embedding.

scripts/question_answering/train_question_answering.py
Ishitori: Done, thanks.

scripts/question_answering/utils.py
Ishitori: Don't know, will talk with Chenguang about that. If yes, then will do it before merging to master.

mli:
Job PR-412/8 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-412/8/index.html

szha:
@Ishitori thanks for contributing the model!
@cgraywang thanks for collaborating on the project. @leezu thanks for the reviewing.

PR #417, opened 2018-11-21
mark tests with network and gpu requirements

### LINKED ISSUE #410: Mark tests with resource needs
We should mark the tests that require network access or GPUs during testing, so that it becomes easier to run tests for people who are on sandboxes. It would also help us run CPU-only tests on machines with better CPU and no GPU, as a way to speed up CI.

szha:
## Description ##
mark tests with network and gpu requirements so that it's easier to test gluonnlp in different environments.

resolves #410 

### Essentials ###
- [x] Code is well-documented

### Changes ###
- [x] mark tests with network and gpu requirements

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,5 @@
+[pytest]
+markers =
+    serial: mark a test that requires more resources to run that are thus only suitable for serial run.
+    remote_required: mark a test that requires internet access.
+    gpu: mark a test that requires GPU.
--- a/scripts/tests/test_question_answering.py
+++ b/scripts/tests/test_question_answering.py
@@ -29,6 +29,7 @@ context_max_length = 256
 
 
 @pytest.mark.serial
+@pytest.mark.remote_required
 def test_transform_to_nd_array():
     dataset = SQuAD(segment='dev', root='tests/data/squad')
     vocab_provider = VocabProvider(dataset)
@@ -41,6 +42,7 @@ def test_transform_to_nd_array():
 
 
 @pytest.mark.serial
+@pytest.mark.remote_required
 def test_data_loader_able_to_read():
     dataset = SQuAD(segment='dev', root='tests/data/squad')
     vocab_provider = VocabProvider(dataset)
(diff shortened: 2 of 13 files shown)


pytest.ini
```diff
@@ -0,0 +1,5 @@
+[pytest]
+markers =
+    serial: mark a test that requires more resources to run that are thus only suitable for serial run.
+    remote_required: mark a test that requires internet access.
+    gpu: mark a test that requires GPU.
```
leezu: How about changing the affected tests (the one to be marked as require gpu) to take a `ctx` (or `gpu_ctx`) argument? Based on that, the markers could be inferred and there would be more control to parallelize GPU heavy tests across multiple GPUs

leezu: (not implying this should be done in this PR, though of course it could)

szha: Sounds like a good idea. How would one skip GPU tests with that setup?

pytest.ini
```diff
@@ -0,0 +1,5 @@
+[pytest]
+markers =
```
szhengac: Where is the code for assigning resources based on markers?

szha: This is not dealt with here. This PR is for making it easier to filter tests when resources are not available.

mli:
Job PR-417/2 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-417/2/index.html

pytest.ini
leezu: Likely something like `https://docs.pytest.org/en/latest/example/markers.html#automatically-adding-markers-based-on-test-names` will work. In short, we need to add `def pytest_collection_modifyitems(items):` to `conftest.py`, which allows us to iterate over all tests (items) and modify them (ie. add a marker)

szha: I see. I’d like to give this more thought. Some immediate concerns are that this may seem somewhat magical and not as explicit, and that naming tests that require network access seem a bit awkward. May I follow this up later?

leezu: Yes, I was not trying to imply this should be done in this PR

szha:
@leezu @szhengac thank you for the timely review

PR #453, opened 2018-12-03
web fixes

szha:
## Description ##
web fixes

### Essentials ###
- [x] Changes are complete (i.e. I finished coding on this PR)

### Changes ###
- [x] fix model zoo link typo for embeddings
- [x] fix edit button

PR #454, opened 2018-12-03
Bump up version

eric-haibin-lin:
## Description ##
As title. 
@cgraywang @leezu @szha @astonzhang


szha:
I'd like to get #453 in before bumping up the version.

leezu:
#450 should be merged first (just requires approval)

PR #456, opened 2018-12-05
Add Chinese, multilingual cased Bert model and Chinese Tokenize 

fierceX:
## Description ##

The Chinese and multilingual cased Bert models were converted according to the script provided by the #449 . In accordance with the #425  added Chinese Tokenize and register vocab and model in GLUONNLP.    

- Add Chinese, multilingual cased Bert model
- Add Chinese Tokenize 

### Essentials ###
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage
- [x] Code is well-documented

### Changes ###
- [ ] Feature1, tests, (and when applicable, API doc)
- [ ] Feature2, tests, (and when applicable, API doc)

## Comments ##
- If this change is a backward incompatible change, why must this change be made.
- Interesting edge cases to note here

PR #464, opened 2018-12-14
[API] Add Bert tokenizer to transforms.py

fierceX:
## Description ##
Add the Bert Tokenizer under the scripts to the transforms https://github.com/dmlc/gluon-nlp/pull/456#discussion_r239242291

### Essentials ###
- [x] Code is well-documented

### Changes ###

## Comments ##


src/gluonnlp/data/transforms.py
```diff
@@ -693,3 +697,283 @@ def __call__(self, sample):
...
+
+    def _whitespace_tokenize(self, text):
+        """Runs basic whitespace cleaning and splitting on a piece of text."""
+        text = text.strip()
+        if not text:
+            return []
+        tokens = text.split()
+        return tokens
+
+    def _convert_to_unicode(self, text):
```
eric-haibin-lin: This looks pretty standalone and we can probably consider making this as a separate Transform class, too. Although I'm ok that if this is not done in this PR

tests/unittest/test_transforms.py
```diff
@@ -150,3 +151,12 @@ def test_sentencepiece_tokenizer_subword_regularization():
...
     assert all(detokenizer(reg_ret[i]) == detext for i in range(len(reg_ret)))
+
+def test_berttokenizer():
+    text = "Introducing Gluon: An Easy-to-Use Programming Interface for Flexible Deep Learning. Gluon: 一个用于深度学习灵活而易于使用的编程接口。"
+    vocab = _load_vocab('wiki_multilingual_cased',vocab=None,root='tests/data/model/')
+    tokenizer = t.BERTTokenizer(vocab=vocab)
+    ret = tokenizer(text)
+    assert isinstance(ret, list)
+    assert len(ret) > 0
+
```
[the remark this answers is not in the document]
eric-haibin-lin: If we also expose BERTBasicTokenizer then we probably need to add corresponding test like the ones in https://github.com/google-research/bert/blob/master/tokenization_test.py#L27-L87

mli:
Job PR-464/8 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-464/8/index.html

src/gluonnlp/data/transforms.py
```diff
@@ -693,3 +696,299 @@ def __call__(self, sample):
             Detokenized text
         """
         return self._processor.DecodePieces(sample)
+
+
+class BasicTokenizer():
+    r"""Runs basic tokenization (punctuation splitting, lower casing, etc.).
+
```
eric-haibin-lin: Let's make it clear in the documentation what tokenization steps are taken, because "BasicTokenizer" by its name is a bit vague:
- convert strings to unicode
- remove invalid chars (e.g. control chars) from the string
- tokenize JCK chars
- ...

src/gluonnlp/data/transforms.py
```diff
@@ -693,3 +696,299 @@ def __call__(self, sample):
...
+    def _tokenizer(self, text):
+        split_tokens = []
+        for token in self.basic_tokenizer(text):
+            for sub_token in self._WordpieceTokenizer(token):
+                split_tokens.append(sub_token)
+
+        return split_tokens
+
+    def _WordpieceTokenizer(self, text):
```
eric-haibin-lin: suggest either make it class `_WordpieceTokenizer` with `__call__` method, or change it to `def _tokenize_wordpiece`

tests/unittest/test_transforms.py
```diff
@@ -150,3 +153,68 @@ def test_sentencepiece_tokenizer_subword_regularization():
...
+def test_berttokenizer():
+
+    # test WordpieceTokenizer
+    vocab_tokens = ["want", "##want", "##ed", "wa", "un", "runn", "##ing"]
+    vocab = Vocab(count_tokens(vocab_tokens), reserved_tokens=[
+                  "[CLS]", "[SEP]"], unknown_token="[UNK]", padding_token=None, bos_token=None, eos_token=None)
+    tokenizer = t.BERTTokenizer(vocab=vocab)
+
+    assert tokenizer._WordpieceTokenizer("") == []
```
eric-haibin-lin: What's the purpose of this assert?

src/gluonnlp/data/transforms.py
```diff
@@ -693,3 +696,299 @@ def __call__(self, sample):
...
         return self._processor.DecodePieces(sample)
+
+
+class BasicTokenizer():
+    r"""Runs basic tokenization (punctuation splitting, lower casing, etc.).
+
+    Parameters
+    ----------
+    do_lower_case : bool, default True
+        If model is uncased, then do_lower_case should be set to True
```
szha: remember that this is a general tokenizer class now and not specific to BERT. the document should clearly define what the tokenizer does, instead of relying on information from model.

src/gluonnlp/data/transforms.py
```diff
@@ -693,3 +696,299 @@ def __call__(self, sample):
...
         """
         return self._processor.DecodePieces(sample)
+
+
+class BasicTokenizer():
+    r"""Runs basic tokenization (punctuation splitting, lower casing, etc.).
+
+    Parameters
+    ----------
+    do_lower_case : bool, default True
```
szha: maybe simplify the name to `lower_case`

src/gluonnlp/data/transforms.py
```diff
@@ -693,3 +696,299 @@ def __call__(self, sample):
...
+
+        output_tokens = _whitespace_tokenize(' '.join(split_tokens))
+        return output_tokens
+
+    def _clean_text(self, text):
+        """Performs invalid character removal and whitespace cleanup on text."""
+        output = []
+        for char in text:
+            cp = ord(char)
+            if cp == 0 or cp == 0xfffd or self._is_control(char):
```
szha: nit: simplify `cp in (0, 0xfffd) or self._is_control(char):`

src/gluonnlp/data/transforms.py
```diff
@@ -693,3 +696,299 @@ def __call__(self, sample):
...
+        if char in [' ', '\t', '\n', '\r']:
+            return True
+        cat = unicodedata.category(char)
+        if cat == 'Zs':
+            return True
+        return False
+
+
+class BERTTokenizer(object):
+    r"""End-to-end tokenziation for BERT models.
```
szha: typo

src/gluonnlp/data/transforms.py
```diff
@@ -693,3 +696,299 @@ def __call__(self, sample):
...
+
+    Parameters
+    ----------
+    vocab : gluonnlp.Vocab or None, default None
+        Vocabulary for the corpus.
+    do_lower_case : bool, default True
+        If model is uncased, then do_lower_case should be set to True
```
szha: same as previous comment

src/gluonnlp/data/transforms.py
szha: If making into a separate, public class, the document format needs to be fixed. GluonNLP adopts the numpy style docstring.

src/gluonnlp/data/utils.py
```diff
@@ -399,3 +400,23 @@ def __call__(self, s):
             List of strings. Obtained by calling s.split(separator).
         """
         return s.split(self._separator)
+
+def _convert_to_unicode(text):
```
szha: instead of having python version inside the function, consider having it outside, so that the version check only happens once.

src/gluonnlp/data/transforms.py
```diff
@@ -693,3 +696,305 @@ def __call__(self, sample):
...
+        Returns
+        -------
+        ret : str
+            Detokenized text
+        """
+        return self._tokenize(sample)
+
+    def _tokenize(self, text):
+        """Tokenizes a piece of text."""
+        text = _convert_to_unicode(text)
```
leezu: Do we need to do encoding conversion inside the tokenizer? If the tokenizer only works for unicode text, why not simply require the user to call it with unicode text? Currently you assume that e.g. `text.decode('utf-8', 'ignore')` is appropriate for every user, but there are valid use cases where another error handler should be chosen, possibly even a custom one https://docs.python.org/3/library/stdtypes.html#bytes.decode

src/gluonnlp/data/utils.py
```diff
@@ -399,3 +400,21 @@ def __call__(self, s):
             List of strings. Obtained by calling s.split(separator).
         """
         return s.split(self._separator)
+
+
+if sys.version_info[0] == 3:
```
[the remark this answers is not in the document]
szha: Good point. Would it be better if this type of version-dependent features be put in the base.py instead?

fierceX:
I removed the code that converted Unicode and indicated in the document a string that must be encoded using Unicode. Please help keep the review going. Besides, why didn't CI pass? @leezu @szha

src/gluonnlp/data/transforms.py
```diff
@@ -731,19 +730,18 @@ def __call__(self, sample):
 
         Parameters
         ----------
-        sample: list(str)
-            The sentence to detokenize
+        sample: str
+            The string to tokenize, the encoding must be Unicode.
```
leezu: Only Python byte strings have an encoding, unicode strings do not. So the current documentation is confusing.

fierceX: Python2 has `str` and `unicode` two types of strings, how do I describe python2 users need to use `unicode` type string?

leezu: In Python2 `str` is byte string and `unicode` is a unicode string. In Python3  `str` is unicode string and `bytes` a byte string. Just replacing "The string to tokenize, the encoding must be Unicode." with "The string to tokenize. Must be unicode." and "sample: str" with "sample: str (unicode for Python 2)" should be fine I guess. Thanks!

mli:
Job PR-464/22 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-464/22/index.html

eric-haibin-lin:
@leezu any other concerns about this PR?

eric-haibin-lin:
@leezu gentle ping

Issue #474, opened 2018-12-21
[BERT] dataset preparation script for pre-training


eric-haibin-lin:
Port https://github.com/google-research/bert/blob/master/create_pretraining_data.py to GluonNLP, preferably written in a way that can be used for processing large dataset

haven-jeon:
I can help on this topic.

eric-haibin-lin:
@haven-jeon I sent an email to `madjakarta@gmail.com` to add you to gluonnlp slack channel. Did you get that?

haven-jeon:
I've done with RecordIO file. 
I'm not sure RecordIO is appropriate...

haven-jeon:
I've done PR #489. 
It is basic conversion from https://github.com/google-research/bert/blob/master/create_pretraining_data.py. 
If it needs documents, Plz let me know where to write.

eric-haibin-lin:
I documented some followup tasks I'm working on in https://github.com/dmlc/gluon-nlp/issues/521

PR #489, opened 2018-12-28
[BERT] dataset preparation script for pre-training

haven-jeon:
## Description ##
- Simple script convertion from https://github.com/google-research/bert/blob/master/create_pretraining_data.py to Gluon(MXRecordIO).

### Essentials ###
- [ ] Code is well-documented

### Changes ###

## Comments ##

### DIFF
--- a/scripts/bert/create_pretraining_data.py
+++ b/scripts/bert/create_pretraining_data.py
@@ -1,0 +1,462 @@
+# coding=utf-8
+# Copyright 2018 The Google AI Language Team Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Create masked LM/next sentence masked_lm Gluon examples for BERT."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import argparse
+import logging
+import pickle
+import io
+import glob
+import collections
+import random
+import mxnet as mx
+import tokenizer as tokenization
+
+logging.getLogger().setLevel(logging.DEBUG)
+
+parser = argparse.ArgumentParser(
+    description='Pretraining data generator for BERT pretraining')
+
+parser.add_argument('--input_file', type=str, default=None, help='input_file')
+
+parser.add_argument(
+    '--output_file',
+    type=str,
+    default=None,
+    help='Output MXRecordIO example file (or comma-separated list of files).')
+
+parser.add_argument(
+    '--vocab_file',
+    type=str,
+    default=None,
+    help='The vocabulary file that the BERT model was trained on.')
+
+parser.add_argument(
+    '--do_lower_case',
+    type=bool,
+    default=False,
+    help='Whether to lower case the input text. Should be True for uncased'
+    'models and False for cased models.')
+
+parser.add_argument(
+    '--max_seq_length', type=int, default=128, help='Maximum sequence length.')
+
+parser.add_argument(
+    '--max_predictions_per_seq',
+    type=int,
+    default=20,
+    help='Maximum number of masked LM predictions per sequence.')
+
+parser.add_argument(
+    '--random_seed',
+    type=int,
+    default=12345,
+    help='Random seed for data generation.')
+
+parser.add_argument(
+    '--dupe_factor',
+    type=int,
+    default=10,
+    help='Number of times to duplicate the input data (with different masks).')
+
+parser.add_argument(
+    '--masked_lm_prob',
+    type=float,
+    default=0.15,
+    help='Masked LM probability.')
+
+parser.add_argument(
+    '--short_seq_prob',
+    type=float,
+    default=0.1,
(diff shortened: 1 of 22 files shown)


haven-jeon:
#474

mli:
Job PR-489/2 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-489/2/index.html

scripts/bert/create_pretraining_data.py
```diff
@@ -0,0 +1,464 @@
...
+    default=0.1,
+    help='Probability of creating sequences which are shorter than the '
+    'maximum length.')
+
+args = parser.parse_args()
+
+logging.info(args)
+
+
+class TrainingInstance:
```
eric-haibin-lin: inherit from object for py2 compatibility

scripts/bert/create_pretraining_data.py
```diff
@@ -0,0 +1,464 @@
...
+
+    writer_index = 0
+
+    total_written = 0
+    for (inst_index, instance) in enumerate(instances):
+        input_ids = tokenizer.convert_tokens_to_ids(instance.tokens)
+        input_mask = [1] * len(input_ids)
+        segment_ids = list(instance.segment_ids)
+        assert len(input_ids) <= max_seq_length
+
```
eric-haibin-lin: For padded data, for consistency let's put 
```
pad = tokenizer.convert_tokens_to_ids('[PAD]')
...
            input_ids.append(pad)
            input_mask.append(pad)
            segment_ids.append(pad)

```

scripts/bert/create_pretraining_data.py
```diff
@@ -0,0 +1,464 @@
...
+            del trunc_tokens[0]
+        else:
+            trunc_tokens.pop()
+
+
+def main():
+    """
+    main function
+    """
+    vocab_obj = pickle.load(open(args.vocab_file, 'rb'))
```
eric-haibin-lin: Would it be better to assume users have some vocab serialized with `file.write(vocab.to_json())`? If so we should use `Vocab.from_json(open(args.vocab_file).read())`? 
This way user can also provide the path to auto-downloaded vocab from pretrained BERT

eric-haibin-lin:
Could you also provide some example code how these recordIO can be read for training?

scripts/bert/create_pretraining_data.py
```diff
@@ -0,0 +1,464 @@
...
+        features['input_mask'] = input_mask
+        features['segment_ids'] = segment_ids
+        features['masked_lm_positions'] = masked_lm_positions
+        features['masked_lm_ids'] = masked_lm_ids
+        features['masked_lm_weights'] = masked_lm_weights
+        features['next_sentence_labels'] = [next_sentence_label]
+
+        example = features
+
+        writers[writer_index].write(pickle.dumps(example))
```
leezu: It might be inefficient to store a RecordIO file of pickled entries? Pickling is very expensive, and is now applied to every sentence individuall during save & load.

If the data fits in memory, consider using a numpy array for serialization. It should allow speeding things up by more than an order of magnitude. When dealing with variable length data, you must concatenate the variable length arrays to a single array and remember the concatenation-points so that you can break up the array again when loading it. See eg https://github.com/dmlc/gluon-nlp/blob/8ec91d8/scripts/machine_translation/dataprocessor.py#L33-L65

What do you think?

mli:
Job PR-489/3 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-489/3/index.html

scripts/bert/create_pretraining_data.py
leezu: (Regarding the code I linked above, it would be simpler to use `numpy.split` instead of the list comprehension and use of `zip` in the load function.)

haven-jeon: Thanks for comments . Also, there is several performance issues to store the duplicated keys for every records.

scripts/bert/create_pretraining_data.py
haven-jeon: Sure!

scripts/bert/create_pretraining_data.py
haven-jeon: It needs to follow json  for object serialization. Thanks.

scripts/bert/create_pretraining_data.py
haven-jeon: Can we be sure padding of  input_mask, segment_ids will be same?

input_mask, segment_ids will have only '0', '1'

This will be correct.

```
pad = tokenizer.convert_tokens_to_ids('[PAD]')
...
            input_ids.append(pad)
            input_mask.append(0)
            segment_ids.append(0)
```

scripts/bert/create_pretraining_data.py
leezu: Possibly https://github.com/dmlc/gluon-nlp/pull/498 can be extended (or similar) to take care of and hide the np.concatenate & np.split

mli:
Job PR-489/4 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-489/4/index.html

scripts/bert/create_pretraining_data.py
eric-haibin-lin: Oh you're right. mask and seg_id should remain 0

mli:
Job PR-489/6 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-489/6/index.html

eric-haibin-lin:
I'm in favor of merging this PR first. I made some incremental enhancement to support numpy output in #504

Issue #476, opened 2018-12-21
[BERT] QA Fine-tuning Example


eric-haibin-lin:
Implement fine-tune example on SQUAD: https://github.com/google-research/bert/blob/master/run_squad.py

eric-haibin-lin:
(@fierceX is leading the effort.)

eric-haibin-lin:
follow up task listed in #502

buechel:
@eric-haibin-lin, it looks like there is a bug between lines 461 through 479 in [here](https://github.com/dmlc/gluon-nlp/blob/master/scripts/bert/finetune_squad.py). Looks like `SQuADTransform` is called twice with identical parameters.

eric-haibin-lin:
Hi @buechel sorry for the confusion. Both`dev_data_transform` and `dev_dataset` uses dev_data as the input, but `dev_data_transform` is processed further in a format suitable for making predictions

PR #493, opened 2018-12-29
[SCRIPT]QA Fine-tuning Example for BERT

fierceX:
## Description ##
add QA Fine-tuning Example for BERT #476 
use Bert Tokenizer from #464   

In squad1.1, use bert_base uncased model, dev_dataset has F1 of 88.52% and EM of 80.98%.(Based on `mxnet-cu90-1.5.0b20190216`).
Ues bert_large uncased model,  dev_dataset has F1 of 90.97% and EM of 84.04%.(Based on `mxnet-cu90-1.5.0b20190216`).
In the `mxnet-cu90-1.5.0b20190112` , use  bert_base uncased model, dev_dataset has F1 of 88.45% and EM of 81.21%.  
Using `mxnet-cu90-1.5.0b20190216` because dropout uses cudnn implementation, training speed is increased by one hour (base model, epochs=2). 
Log in https://github.com/dmlc/web-data/pull/161

In squad2.0 use bert_large uncased model,  `null_score_diff_threshold=-2.0`,The results of the dev data set are as follows((Based on `mxnet-cu90-1.5.0b20190216`):
```json
{
  "exact": 77.958392992504,
  "f1": 81.02012658815627,
  "total": 11873,
  "HasAns_exact": 73.3974358974359,
  "HasAns_f1": 79.52968336389662,
  "HasAns_total": 5928,
  "NoAns_exact": 82.50630782169891,
  "NoAns_f1": 82.50630782169891,
  "NoAns_total": 5945
}
```
Log in https://github.com/dmlc/web-data/pull/164

optimizer is adam, lr=3e-5, beta1=0.9, beta2=0.999, epsilon=1e-08.（original repo optimizer  is adamw, lr=3e-5,wd=0.01,beta_1=0.9,beta_2=0.999,epsilon=1e-6,)

### Essentials ###
- [x] Code is well-documented

## Comments ##


eric-haibin-lin:
This is great contribution. I'll take a look later when I get time. Do you plan to write a tutorial for finetuning on the QA task?

fierceX:
@eric-haibin-lin  Ok, I'll try to write a tutorial when I have time.

scripts/bert_SQuAD/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# pylint: disable=wildcard-import
+"""BERT example."""
+from . import bert, dataset, evaluate
```
eric-haibin-lin: Could you add a rst file instructing the command used to reproduce the result?

scripts/bert_SQuAD/bert.py
```diff
@@ -0,0 +1,90 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+"""BERT models."""
+
+__all__ = ['BERTSquad', 'BERTloss']
```
eric-haibin-lin: nit: BERTloss -> BERTLoss

scripts/bert_SQuAD/bert.py
```diff
@@ -0,0 +1,90 @@
...
+        See document of `mx.gluon.Block`.
+    params : ParameterDict or None
+        See document of `mx.gluon.Block`.
+    """
+
+    def __init__(self, bert, prefix=None, params=None):
+        super(BERTSquad, self).__init__(prefix=prefix, params=params)
+        self.bert = bert
+        with self.name_scope():
+            self.Dense = nn.Dense(units=2, flatten=False)
```
eric-haibin-lin: use lower case to indicate members. Also better to use a more meaningful name such as `self.classifier`

scripts/bert_SQuAD/bert.py
```diff
@@ -0,0 +1,90 @@
...
+class BERTloss(Loss):
+    """Loss for SQuAD task with BERT.
+
+    """
+
+    def __init__(self, weight=None, batch_axis=0, **kwargs):
+        super(BERTloss, self).__init__(weight=None, batch_axis=0, **kwargs)
+        self.loss = loss.SoftmaxCELoss()
+
+    def hybrid_forward(self, F, pred, label): # pylint: disable=arguments-differ
```
eric-haibin-lin: Please document input/output shapes

scripts/bert_SQuAD/dataset.py
```diff
@@ -0,0 +1,493 @@
...
+                        prev_is_whitespace = True
+                    else:
+                        if prev_is_whitespace:
+                            doc_tokens.append(c)
+                        else:
+                            doc_tokens[-1] += c
+                        prev_is_whitespace = False
+                    char_to_word_offset.append(len(doc_tokens) - 1)
+
+                for qa in paragraph['qas']:
```
eric-haibin-lin: It looks these can also be built as "Transforms" on top of the SQuAD dataset. https://github.com/dmlc/gluon-nlp/blob/master/src/gluonnlp/data/question_answering.py#L36 What is the reason the SQuAD dataset API is not reused?

scripts/bert_SQuAD/dataset.py
```diff
@@ -0,0 +1,493 @@
...
+                        start_position=start_position,
+                        end_position=end_position,
+                        is_impossible=is_impossible)
+                    examples.append(example)
+
+                    example_id += 1
+
+        return examples
+
+    def whitespace_tokenize(self, text):
```
eric-haibin-lin: Reuse https://github.com/dmlc/gluon-nlp/blob/master/scripts/bert/tokenizer.py#L91

scripts/bert_SQuAD/dataset.py
```diff
@@ -0,0 +1,493 @@
...
+        text = text.strip()
+        if not text:
+            return []
+        tokens = text.split()
+        return tokens
+
+
+class SQuADTransform(object):
+    """Dataset Transformation for BERT-style QA.
+
```
eric-haibin-lin: Need to document down the transformations in the API doc. Let's try to make these code reusable and eventually merge back as gluonnlp APIs

scripts/bert_SQuAD/dataset.py
```diff
@@ -0,0 +1,493 @@
...
+    Parameters
+    ----------
+    tokenizer : BERTTokenizer.
+        Tokenizer for the sentences.
+    labels : list of int.
+        List of all label ids for the classification task.
+    max_seq_length : int, default 384
+        Maximum sequence length of the sentences.
+    doc_stride : int, default 128
+        pass
```
eric-haibin-lin: Need to explain what doc_stride is for

scripts/bert_SQuAD/dataset.py
```diff
@@ -0,0 +1,493 @@
...
+
+            input_ids = self.tokenizer.convert_tokens_to_ids(tokens)
+
+            # The mask has 1 for real tokens and 0 for padding tokens. Only real
+            # tokens are attended to.
+            valid_length = len(input_ids)
+
+            # Zero-pad up to the sequence length.
+            while len(input_ids) < self.max_seq_length:
+                input_ids.append(0)
```
eric-haibin-lin: use `vocab['[PAD]']` instead of 0

scripts/bert_SQuAD/evaluate.py
```diff
@@ -0,0 +1,422 @@
...
+
+def metric_max_over_ground_truths(metric_fn, prediction, ground_truths):
+    scores_for_ground_truths = []
+    for ground_truth in ground_truths:
+        score = metric_fn(prediction, ground_truth)
+        scores_for_ground_truths.append(score)
+    return max(scores_for_ground_truths)
+
+
+def Get_F1_EM(dataset_file, predict_data):
```
eric-haibin-lin: style: get_F1_EM

scripts/bert_SQuAD/bert.py
```diff
@@ -0,0 +1,104 @@
...
+        See document of `mx.gluon.Block`.
+    params : ParameterDict or None
+        See document of `mx.gluon.Block`.
+    """
+
+    def __init__(self, bert, prefix=None, params=None):
+        super(BertForQA, self).__init__(prefix=prefix, params=params)
+        self.bert = bert
+        with self.name_scope():
+            self.classifier = nn.HybridSequential(prefix=prefix)
```
eric-haibin-lin: No need to hybrid sequential if it only contains a dense layer

scripts/bert_SQuAD/finetune_squad.py
```diff
@@ -0,0 +1,435 @@
...
+    net.classifier.initialize(init=mx.init.Normal(0.02), ctx=ctx)
+else:
+    net.load_parameters(model_parameters, ctx=ctx)
+net.hybridize(static_alloc=True)
+
+loss_function = BertForQALoss()
+loss_function.hybridize(static_alloc=True)
+
+
+def Train():
```
eric-haibin-lin: style: def train()

scripts/bert_SQuAD/finetune_squad.py
```diff
@@ -0,0 +1,435 @@
...
+                             .format(epoch_id, batch_id, len(train_dataloader),
+                                     step_loss / log_interval,
+                                     trainer.learning_rate, toc - tic))
+                tic = time.time()
+                step_loss = 0.0
+
+    net.save_parameters(os.path.join(output_dir, 'net_parameters'))
+
+
+def Evaluate():
```
eric-haibin-lin: style: def evaluate()

scripts/bert_SQuAD/__init__.py
eric-haibin-lin: Let's move the example to /scripts/bert/

scripts/bert_SQuAD/dataset.py
fierceX: Please help review the documentation comments.

zhiheng-huang:
Nice work. The current code takes around 70 mins to train one epoch while other DL frameworks reported 30 mins per epoch (or even less time on half precision). Any comment on this? Also, the current code only supports one GPU, any plan to support multi-gpu soon?

szha:
@zhiheng-huang thanks for helping measure the performance. Would you mind sharing the settings and tools you used for the measure, so that other contributors can use the same setting to try tracking down the performance issue?

Regardless of whether it would be added in this PR, Multi-GPU should be a simple extension to the script and can be added either in this PR or soon in the future once it's in good shape.

eric-haibin-lin:
@zhiheng-huang we did find inefficient implementation of Dropout operator which causes the major slowdown. @szha is working on an efficient version which uses CUDNN. 
Regarding multi-GPU, it would be a simple extension following standard practice with a few lines of code.

fierceX:
user squad api #551

mli:
Job PR-493/44 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-493/44/index.html

mli:
Job PR-493/45 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-493/45/index.html

mli:
Job PR-493/47 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-493/47/index.html

mli:
Job PR-493/48 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-493/48/index.html

scripts/bert/qa_dataset.py
```diff
@@ -0,0 +1,516 @@
...
+
+    """
+    worker_fn = partial(_worker_fn, transform=transform)
+    start = time.time()
+
+    with mp.Pool(num_workers) as pool:
+        dataset_transform = []
+        for data in pool.map(worker_fn, dataset):
+            if data:
+                dataset_transform.extend(data)
```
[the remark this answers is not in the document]
fierceX: `SQuADTransform` returns an array of multiple features, so you need to use `extend`

scripts/bert/qa_dataset.py
```diff
@@ -0,0 +1,516 @@
...
+        The number of multiprocessing workers to use for data preprocessing.
+
+    """
+    worker_fn = partial(_worker_fn, transform=transform)
+    start = time.time()
+
+    with mp.Pool(num_workers) as pool:
+        dataset_transform = []
+        for data in pool.map(worker_fn, dataset):
+            if data:
```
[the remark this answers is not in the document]
fierceX: In the original repro there is some code to skip some questions. After I use the squad api, I also need to follow this. But in my test squad1.1 did not skip the questions. But I think it still needs to be retained.

scripts/bert/finetune_squad.py
```diff
@@ -0,0 +1,456 @@
...
+                    default=30,
+                    help='The maximum length of an answer that can be generated. This is needed '
+                    'because the start and end predictions are not conditioned on one another.'
+                    ' default is 30')
+
+parser.add_argument('--version_2',
+                    action='store_true',
+                    help='SQuAD examples whether contain some that do not have an answer.')
+
+parser.add_argument('--null_score_diff_threshold',
```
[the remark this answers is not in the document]
fierceX: There is no code to automatically select this parameter in the original repo, they suggest that the parameter can be selected between -1 and -5.

scripts/bert/finetune_squad.py
```diff
@@ -0,0 +1,456 @@
...
+
+
+version_2 = args.version_2
+null_score_diff_threshold = args.null_score_diff_threshold
+
+max_seq_length = args.max_seq_length
+doc_stride = args.doc_stride
+max_query_length = args.max_query_length
+n_best_size = args.n_best_size
+max_answer_length = args.max_answer_length
```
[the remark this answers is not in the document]
fierceX: Mainly because if someone needs to run the code on `jupyter`, it is not convenient to use args to modify them all.

mli:
Job PR-493/53 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-493/53/index.html

scripts/bert/finetune_squad.py
```diff
@@ -0,0 +1,459 @@
...
+                    default=30,
+                    help='The maximum length of an answer that can be generated. This is needed '
+                    'because the start and end predictions are not conditioned on one another.'
+                    ' default is 30')
+
+parser.add_argument('--version_2',
+                    action='store_true',
+                    help='SQuAD examples whether contain some that do not have an answer.')
+
+parser.add_argument('--null_score_diff_threshold',
```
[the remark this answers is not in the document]
fierceX: Because the original repo indicates that the parameter can be selected between -1 and -5. Need to be consistent with the original repo code?

scripts/bert/finetune_squad.py
```diff
@@ -0,0 +1,459 @@
...
+                    type=int,
+                    default=12,
+                    help='Batch size. Number of examples per gpu in a minibatch. default is 12')
+
+parser.add_argument('--test_batch_size',
+                    type=int,
+                    default=24,
+                    help='Test batch size. default is 24')
+
+parser.add_argument('--optimizer',
```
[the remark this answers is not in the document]
fierceX: Because bertadam needs mxnet 1.5 or above, I think it is reasonable to keep adam now.

scripts/bert/finetune_squad.py
```diff
@@ -0,0 +1,459 @@
...
+                    type=int,
+                    default=2,
+                    help='number of epochs, default is 2')
+
+parser.add_argument('--batch_size',
+                    type=int,
+                    default=12,
+                    help='Batch size. Number of examples per gpu in a minibatch. default is 12')
+
+parser.add_argument('--test_batch_size',
```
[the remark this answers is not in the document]
fierceX: test_batch_size does not affect the effect, on the contrary, improve test_batch_size can improve the verification speed

scripts/bert/finetune_squad.py
```diff
@@ -0,0 +1,459 @@
...
+log.addHandler(console)
+log.addHandler(fh)
+
+log.info(args)
+
+model_name = args.model
+dataset_name = args.dataset_name
+only_predict = args.only_predict
+model_parameters = args.model_parameters
+lower = args.cased
```
bikestra:
```suggestion
lower = args.uncased
```

This seems like a bug.

scripts/bert/finetune_squad.py
```diff
@@ -0,0 +1,459 @@
...
+
+parser.add_argument('--lr',
+                    type=float,
+                    default=3e-5,
+                    help='Initial learning rate. default is 3e-5')
+
+parser.add_argument('--warmup_ratio',
+                    type=float,
+                    default=0.1,
+                    help='ratio of warmup steps used in NOAM\'s stepsize schedule. default is 0.1')
```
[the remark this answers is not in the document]
eric-haibin-lin: Yes

scripts/bert/finetune_squad.py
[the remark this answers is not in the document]
cgraywang: agree with you that we do not necessarily need to follow the original implementation, but only when our new hyper-parameter could generate a better result. Have you verified that `null_score_diff_threshold=-2` generate better results than `=0`?

scripts/bert/finetune_squad.py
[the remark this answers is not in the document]
cgraywang: yes or no, it depends on whether the length of the dev dataset is dividable by the `test_batch_size` and whether the last batch is kept or discarded. could you please verify? thanks.

scripts/bert/finetune_squad.py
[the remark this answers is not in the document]
cgraywang: got it, I noticed you have an assert in the script, but it would be better if we could also add this information in the parameter description.

eric-haibin-lin:
Could you sync with master? The failed link check is fixed on the master branch

mli:
Job PR-493/58 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-493/58/index.html

scripts/bert/finetune_squad.py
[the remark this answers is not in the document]
fierceX: In the code, I kept the last batch

scripts/bert/finetune_squad.py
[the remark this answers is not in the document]
fierceX: In squad2.0, I tested null_score_diff_threshold=-2 better than null_score_diff_threshold=0. In the original repo description, the recommended null_score_diff_threshold range is between -1 and -5.

scripts/bert/finetune_squad.py
```diff
@@ -0,0 +1,459 @@
...
+                    default='bertadam',
+                    help='optimization algorithm. default is bertadam')
+
+parser.add_argument('--accumulate',
+                    type=int,
+                    default=None,
+                    help='The number of batches for '
+                    'gradients accumulation to simulate large batch size. Default is None')
+
+parser.add_argument('--lr',
```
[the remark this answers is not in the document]
fierceX: Because the large model in the original paper is not reproduced, only the results of the pre-training in the original repo are reproduced, whether the default value of the parameter is necessary to be consistent with the original paper, or the parameter value of the original repo that can be reproduced is used.

scripts/bert/index.rst
```diff
@@ -37,3 +37,32 @@ Use the following command to fine-tune the BERT model for classification on the
...
+
+    $ python finetune_squad.py --optimizer adam --batch_size 12 --lr 3e-5 --epochs 2 --gpu
+
+If you are using less than 12G of GPU memory, you can use the following command to achieve a similar effect. But need Mxnet>1.5.0
+
+Note that this will require approximately no more than 8G of GPU memory. If your GPU memory is too small, you can adjust **accumulate** and **batch_size**.
+
+.. code-block:: console
+
+    $ python finetune_squad.py --optimizer bertadam --accumulate 2 --batch_size 6 --lr 3e-5 --epochs 2 --gpu
```
eric-haibin-lin: Why is bertadam used in this line and adam is used in line 53? Is that a typo?

eric-haibin-lin: Grad accumulation is supposed to work irrespective of the choice of optimizer. What did you observe when adam is used with grad accumulation?

fierceX: I have not tested how adam works with gradient accumulation. But the above results are obtained with adam

mli:
Job PR-493/46 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-493/46/index.html

scripts/bert/bert_qa_dataset.py
```diff
@@ -0,0 +1,516 @@
...
+            question_text=question_text,
+            doc_tokens=doc_tokens,
+            example_id=example_id,
+            orig_answer_text=orig_answer_text,
+            start_position=start_position,
+            end_position=end_position,
+            is_impossible=is_impossible)
+        return example
+
+    def _transform(self, *record):
```
cgraywang: should we change this as the default `__call__` function? since when calling `SQuADTransform `, it only uses this function. we thus could directly return,

scripts/bert/bert_qa_dataset.py
```diff
@@ -0,0 +1,516 @@
...
+            features.append(feature)
+
+        return features
+
+
+def bert_qa_batchify_fn(data):
+    """Collate data into batch."""
+    def batchify_fn(data):
+        data = np.asarray(data)
+        return nd.array(data, dtype=data.dtype, ctx=context.Context('cpu_shared', 0))
```
cgraywang: why are we using `ctx=context.Context('cpu_shared', 0)`?

scripts/bert/finetune_squad.py
```diff
@@ -0,0 +1,475 @@
...
+            if batch_id % accumulate == 0:
+                net.collect_params().zero_grad()
+                step_num += 1
+        else:
+            step_num += 1
+        # learning rate schedule
+        if step_num < num_warmup_steps:
+            new_lr = lr * step_num / num_warmup_steps
+        else:
+            offset = (step_num - num_warmup_steps) * lr / \
```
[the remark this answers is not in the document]
cgraywang: @eric-haibin-lin we probably need to address this.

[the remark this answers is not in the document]
eric-haibin-lin: I didn't realize that our implementation is inconsistent with the original repo. I suggest we add a comment here that our learning rate schedule is a bit different from the original repo, so that others are aware. We probably don't have to change it and re-train everything, since we have already reproduced the result.

scripts/bert/bert_qa_evaluate.py
```diff
@@ -0,0 +1,409 @@
...
+
+def metric_max_over_ground_truths(metric_fn, prediction, ground_truths):
+    scores_for_ground_truths = []
+    for ground_truth in ground_truths:
+        score = metric_fn(prediction, ground_truth)
+        scores_for_ground_truths.append(score)
+    return max(scores_for_ground_truths)
+
+
+def get_F1_EM(dataset, predict_data):
```
cgraywang: I forget to mention that, as the evaluation code is adopted from official evaluation script, it would be better if we can cite their code: https://worksheets.codalab.org/rest/bundles/0xbcd57bee090b421c982906709c8c27e1/contents/blob/

scripts/bert/finetune_squad.py
```diff
@@ -0,0 +1,475 @@
...
+                                 step_loss / log_interval,
+                                 trainer.learning_rate, toc - tic, log_num/(toc - tic)))
+                tic = time.time()
+                step_loss = 0.0
+                log_num = 0
+        epoch_toc = time.time()
+        log.info('Time cost={:.2f} s, Thoughput={:.2f} samples/s'.format(
+            epoch_toc - epoch_tic, total_num/(epoch_toc - epoch_tic)))
+
+    net.save_parameters(os.path.join(output_dir, 'net_parameters'))
```
eric-haibin-lin: We usually use xxx.params for gluon params. Change to `net.params`?

scripts/bert/bert_qa_dataset.py
```diff
@@ -0,0 +1,516 @@
...
+        self.doc_span_index = doc_span_index
+        self.tokens = tokens
+        self.token_to_orig_map = token_to_orig_map
+        self.token_is_max_context = token_is_max_context
+        self.input_ids = input_ids
+        self.valid_length = valid_length
+        self.segment_ids = segment_ids
+        self.start_position = start_position
+        self.end_position = end_position
+        self.is_impossible = is_impossible
```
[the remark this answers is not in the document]
fierceX: I think using `is_impossible` will be more readable

scripts/bert/bert_qa_dataset.py
fierceX: The result of the _transform function is used when calculating the answer.

scripts/bert/bert_qa_dataset.py
fierceX: This is my reference to the implementation of gluon in mxnet

mli:
Job PR-493/62 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-493/62/index.html

scripts/bert/bert_qa_evaluate.py
fierceX: When I was looking for the `evaluate-v1.1` script, I didn't find the official script.
But there are scripts from https://github.com/allenai/bi-att-flow/blob/master/squad/evaluate-v1.1.py in the original repo.
For the sake of convenience, I merged the script code into `bert_qa_evaluate.py`.
Is it necessary to provide the original repo script address or official script address.

mli:
Job PR-493/64 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-493/64/index.html

fierceX:
Because using gluonnlp.data.batchify.Pad to dynamically fill the batch length will cause the GPU memory usage to increase, so I change the code to the pre-filled maximum length. #590

mli:
Job PR-493/65 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-493/65/index.html

mli:
Job PR-493/66 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-493/66/index.html

scripts/bert/finetune_squad.py
```diff
@@ -0,0 +1,485 @@
...
+        doc_stride=doc_stride,
+        max_query_length=max_query_length,
+        is_pad=True,
+        is_training=True))
+    log.info('The number of examples after preprocessing:{}'.format(
+        len(train_data_transform)))
+
+    train_dataloader = mx.gluon.data.DataLoader(
+        train_data_transform, batch_size=batch_size,
+        batchify_fn=batchify_fn, num_workers=4, shuffle=True)
```
kenjewu: #590 . Use bucketing here to improve performance, such as [nlp.data.sampler.FixedBucketSampler](https://gluon-nlp.mxnet.io/api/notes/data_api.html#bucketing-and-dataloader).

scripts/bert/index.rst
```diff
@@ -74,4 +77,15 @@ Should produce an output like this. Explain that the F1 score on the dev dataset
...
+SQuAD 2.0
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you are pre-training on the SQuAD2.0 dataset, you need to specify the parameter **version_2** and specify the parameter **null_score_diff_threshold** (Typical values are between -1.0 and -5.0).
+
+Use the following command to fine tune the BERT model of the SQuAD2.0 dataset and generate predictions.json, nbest_predictions.json, and null_odds.json.
+
+.. code-block:: console
+
+    $ python finetune_squad.py --optimizer adam --batch_size 12 --lr 3e-5 --epochs 2 --null_score_diff_threshold -2.0 --gpu --version_2
```
cgraywang: Forgot to append the results for SQuAD 2.0?

scripts/bert/finetune_squad.py
cgraywang: Agree, use `batch_sampler=FixedBucketSampler` could be helpful.

scripts/bert/bert_qa_dataset.py
```diff
@@ -76,7 +77,8 @@ def preprocess_dataset(dataset, transform, num_workers=8):
             if data:
                 dataset_transform.extend(data)
 
-        dataset = SimpleDataset(dataset_transform)
+        dataset = SimpleDataset(dataset_transform).transform(
+            lambda x: (x[0], x[1], x[2], x[3], x[4], x[5]))
```
cgraywang: when SQuAD=2.0, would this be `lambda x: (x[0], x[1], x[2], x[3], x[4], x[5], x[6]))`?

fierceX: No, `x[0], x[1], x[2], x[3], x[4], x[5]` are `example_ids, inputs, token_types, valid_length, start_label, end_label`

scripts/bert/index.rst
fierceX: The original repo did not provide the results of squad2.0, do we need to provide it?

scripts/bert/finetune_squad.py
fierceX: After using `FixedBucketSampler`, GPU memory will also increase, and will gradually increase to the same size as before.

scripts/bert/finetune_squad.py
```diff
@@ -404,23 +411,30 @@ def evaluate():
...
+            is_training=False))
     log.info('The number of examples after preprocessing:{}'.format(
         len(dev_data_transform)))
 
+    batch_sampler = nlp.data.sampler.FixedBucketSampler(data_dev_len,
+                                                        batch_size=test_batch_size,
+                                                        num_buckets=10,
+                                                        ratio=0,
+                                                        shuffle=True)
+
```
kenjewu: Because the validation data is generally small, the batch sampler is usually not used for validation. Even if needed, it is better to set ` shuffle ` to ` False `.

scripts/bert/index.rst
kenjewu: You can check whether the length of data is reasonable by outputting ` batch_sampler.stats() `. If have a lot of long samples, you may still need to clip them.

mli:
Job PR-493/67 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-493/67/index.html

scripts/bert/index.rst
cgraywang: I think it would be great if we have the results for 2.0. Is there any blocker?

scripts/bert/finetune_squad.py
eric-haibin-lin: Good point. For val data please set shuffle to False

fierceX:
After using `FixedBucketSampler`, the GPU has about 15G, but there will be a chance of OOM, so I think it is appropriate to fix the length to the maximum length.
Using the `FixedBucketSampler` does increase the training speed significantly, saving about an hour and a half (the original total of about two and a half hours).

mli:
Job PR-493/68 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-493/68/index.html

scripts/bert/index.rst
eric-haibin-lin: @cgraywang The reported squad 2.0 test f1 is using BERT LARGE and uses lots of memory. Also the official repo didn't report the dev f1 on squad 2.0. Maybe we can set this as a stretch goal for this PR.

scripts/bert/finetune_squad.py
```diff
@@ -0,0 +1,487 @@
...
+                log_num += len(inputs)
+                total_num += len(inputs)
+
+                out = net(inputs.astype('float32').as_in_context(ctx),
+                          token_types.astype('float32').as_in_context(ctx),
+                          valid_length.astype('float32').as_in_context(ctx))
+
+                ls = loss_function(out, [
+                    start_label.astype('float32').as_in_context(ctx),
+                    end_label.astype('float32').as_in_context(ctx)]).mean()
```
eric-haibin-lin:
```
if accumulate:
    ls = ls / accumulate
```

scripts/bert/finetune_squad.py
```diff
@@ -0,0 +1,487 @@
...
+            ls.backward()
+            # update
+            if not accumulate or (batch_id + 1) % accumulate == 0:
+                trainer.allreduce_grads()
+                nlp.utils.clip_grad_global_norm(params, 1)
+                trainer.update(accumulate if accumulate else 1)
```
eric-haibin-lin: trainer.update(1)

mli:
Job PR-493/70 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-493/70/index.html

mli:
Job PR-493/71 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-493/71/index.html

scripts/bert/index.rst
cgraywang: I think we can run the experiments using BERT-base. It would be good to give the users the expectations if they want to use squad 2.0 and want to know whether they are generating reasonable results. Another option could be removing the squad 2.0 option in this script.

cgraywang:
@fierceX , I think `FixedBucketSampler` could be a better option, and it is also a little bit weired why use it will cost significant more memory compared purely PAD operation. Could you also try other options in `FixedBucketSampler`? e.g., setting bucket_scheme=ExpWidthBucket and use_average_length=True.

scripts/bert/index.rst
fierceX: Ok, I will give the results of the Large model and the training log later.

fierceX:
@cgraywang I also think it is better to use FixedBucketSampler, but now the output length is not fixed after using FixedBucketSampler. This will result in a significant increase in memory usage and a high probability of OOM. This PR tries to reproduce the results of the original repo, so I don't think I will use FixedBucketSampler for the time being.

cgraywang:
@fierceX it is interesting that there is a result drop since the overall data distribution stays the same. How much are the results drop? Besides, probably a simple revision to the previous FixedBucketSampler would work. If the results drop is not significant, it worth trying those options in the previous comment. Thanks.

scripts/bert/index.rst
```diff
@@ -45,4 +45,47 @@ It gets validation accuracy of `88.7% <https://raw.githubusercontent.com/dmlc/we
...
+SQuAD 1.1
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Use the following command to fine-tune the BERT model for SQuAD1.1 dataset.
+
+Note that this will require more than 12G of GPU memory.
+ 
+.. code-block:: console
+
+    $ python finetune_squad.py --optimizer adam --batch_size 12 --lr 3e-5 --epochs 2 --gpu
```
eric-haibin-lin: Could you also add the instructions on how to download the squad dataset and how to run evaluation?

scripts/bert/finetune_squad.py
```diff
@@ -0,0 +1,492 @@
...
+                                 'We fine-tune the BERT model on SQuAD dataset.')
+
+parser.add_argument('--only_predict',
+                    action='store_true',
+                    help='Whether to predict only.')
+
+parser.add_argument('--model_parameters',
+                    type=str,
+                    default=None,
+                    help='Model parameter file')
```
eric-haibin-lin: Could you also provide the option to load pretrained bert parameters like in https://github.com/dmlc/gluon-nlp/pull/571/ ?

scripts/bert/bert_qa_dataset.py
```diff
@@ -0,0 +1,505 @@
...
+    transform: callable
+        Transformer function.
+    num_workers: int, default 8
+        The number of multiprocessing workers to use for data preprocessing.
+
+    """
+    worker_fn = partial(_worker_fn, transform=transform)
+    start = time.time()
+
+    with mp.Pool(num_workers) as pool:
```
eric-haibin-lin: I got this error when running the script in py2:
```
Traceback (most recent call last):
  File "finetune_squad.py", line 490, in <module>
    train()
  File "finetune_squad.py", line 288, in train
    is_training=True))
  File "/home/ubuntu/qa-nlp/scripts/bert/bert_qa_dataset.py", line 72, in preprocess_dataset
    with mp.Pool(num_workers) as pool:
AttributeError: __exit__
```
Can we avoid using the `with` stmt for py2 compatibility?

mli:
Job PR-493/75 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-493/75/index.html

mli:
Job PR-493/78 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-493/78/index.html

mli:
Job PR-493/79 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-493/79/index.html

fierceX:
@cgraywang Because there is a chance of OOM, I didn't test too many different seeds, but from the few tests, the result dropped by about 1%. This may be the reason why shuffle and FixedBucketSampler are different after using FixedBucketSampler . It is also possible that no suitable seed has been found.
The main reason why RP does not use FixedBucketSampler here is that OOM will occur.

eric-haibin-lin:
Merged. Very awesome work. Thanks everyone

PR #507, opened 2019-01-03
[Feature] Support custom sampler in SimpleDatasetStream

eric-haibin-lin:
## Description ##
Add support for custom sampler in SimpleDatasetStream, so that SplitSampler can be passed as an argument. 

### Essentials ###
- [x] Code is well-documented

### Changes ###

## Comments ##

### DIFF
--- a/src/gluonnlp/data/stream.py
+++ b/src/gluonnlp/data/stream.py
@@ -35,7 +35,7 @@ import traceback
 import numpy as np
 
 import mxnet as mx
-from mxnet.gluon.data import RandomSampler, SequentialSampler
+from mxnet.gluon.data import RandomSampler, SequentialSampler, Sampler
 
 try:
     import Queue as queue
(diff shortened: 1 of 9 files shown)


mli:
Job PR-507/1 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-507/1/index.html

src/gluonnlp/data/stream.py
```diff
@@ -170,8 +170,8 @@ class SimpleDatasetStream(DatasetStream):
         passed to this class.
     file_pattern: str
         Path to the input text files.
-    file_sampler : str, {'sequential', 'random'}, defaults to 'random'
-        The sampler used to sample a file.
+    file_sampler : str or gluon.data.Sampler, defaults to 'random'
```
szha: I wonder if "file" is the right unit of the chunk of data. Each chunk in a dataset stream should be a dataset, so the API design should reflect that.

eric-haibin-lin: I'm not sure what is the best way to define a "chunk". Currently Corpus/Textline dataset both take "file" as an argument to construct a dataset.

eric-haibin-lin:
I'll merge the sampler change for now. If we need to refine the definition for "chunk", we shall do it in a separate PR

Issue #497, opened 2018-12-30
[BERT] BERT Tokenizer PR followups 


eric-haibin-lin:
- [ ] Add a de-tokenizer API which returns the original string representation given tokens
- [ ] Replace existing tokenization code in `script/bert` with the new tokenizer API

eric-haibin-lin:
ping @kenjewu if you are interested

kenjewu:
ok, i will try to finish it and #492 .

PR #509, opened 2019-01-04
[API]BERTVocab

kenjewu:
## Description ##
Added BERTVocab and replaced some string literals in existing bert examples. #492

### Essentials ###
- [x] PR's title starts with a category (e.g. [BUGFIX], [MODEL], [TUTORIAL], [FEATURE], [DOC], etc)
- [x] Changes are complete (i.e. I finished coding on this PR)
- [x] All changes have test coverage
- [x] Code is well-documented

### Changes ###
replaced some string literals in existing bert examples.

## Comments ##

PR #526, opened 2019-01-05
[API] Add Transforms for BERT

kenjewu:
## Description ##
#497 , Organized some APIs about BERT, such as BERTTransform, BERTClassificationTransform, BERTTokenizer, etc.  And replace the relevant code in scripts/bert with the relevant API to make the code more concise.
The related API documentation has not been updated to the doc yet, and will be updated after all API tests have been completed.

### Essentials ###
- [x] Code is well-documented

### Changes ###
added some new APIs about BERT and  removed some files in scripts/bert.


src/gluonnlp/data/transforms.py
```diff
@@ -988,10 +1004,261 @@ def convert_tokens_to_ids(self, tokens):
...
+        # (and position vector). This is not *strictly* necessary since
+        # the [SEP] token unambiguously separates the sequences, but it makes
+        # it easier for the model to learn the concept of sequences.
+
+        # For classification tasks, the first vector (corresponding to [CLS]) is
+        # used as as the "sentence vector". Note that this only makes sense because
+        # the entire model is fine-tuned.
+        tokens = []
+        segment_ids = []
+        tokens.append('[CLS]')
```
eric-haibin-lin: Reminder to replace '[CLS]' with the one from BERTVocab once https://github.com/dmlc/gluon-nlp/pull/509 is merged

src/gluonnlp/data/transforms.py
```diff
@@ -988,10 +1004,261 @@ def convert_tokens_to_ids(self, tokens):
...
+            return text.decode('utf-8', 'ignore')
+        elif isinstance(text, unicode):  # noqa: F821
+            return text
+        else:
+            raise ValueError('Unsupported string type: %s' % (type(text)))
+    else:
+        raise ValueError('Not running on Python2 or Python 3?')
+
+
+def _truncate_seq_pair(tokens_a, tokens_b, max_length):
```
eric-haibin-lin: This is specific for BERTClassificationTransform. Put it in that class

src/gluonnlp/data/transforms.py
```diff
@@ -988,10 +1004,261 @@ def convert_tokens_to_ids(self, tokens):
         return self.vocab.to_indices(tokens)
 
 
+class BERTTransform(object):
```
eric-haibin-lin: BERTSentenceTransform

src/gluonnlp/data/transforms.py
```diff
@@ -988,10 +1004,261 @@ def convert_tokens_to_ids(self, tokens):
...
+            padding_length = self._max_seq_length - valid_length
+            # use padding tokens for the rest
+            input_ids.extend([self._tokenizer.vocab['[PAD]']] * padding_length)
+            segment_ids.extend([self._tokenizer.vocab['[PAD]']] * padding_length)
+
+        return np.array(input_ids, dtype='int32'), np.array(valid_length, dtype='int32'),\
+            np.array(segment_ids, dtype='int32')
+
+
+class BERTClassificationTransform(object):
```
eric-haibin-lin: Can we unify the interface for classification and regression? We can have a reusable transform for both classification and regression. 
The transform should take a label_dtype argument (default to 'float32'). For classification users can set dtype to 'int32' explicitly.

src/gluonnlp/data/transforms.py
```diff
@@ -988,10 +1004,261 @@ def convert_tokens_to_ids(self, tokens):
...
+        Whether to transform sentences or sentence pairs.
+    """
+
+    def __init__(self, tokenizer, labels, max_seq_length, pad=True, pair=True):
+        self._label_map = {}
+        for (i, label) in enumerate(labels):
+            self._label_map[label] = i
+        self._bert_xform = BERTTransform(tokenizer, max_seq_length, pad=pad, pair=pair)
+
+    def __call__(self, line):
```
eric-haibin-lin: Do you mind also adding `__call__` to "special members" so that the doc for `__call__` would display? https://github.com/dmlc/gluon-nlp/blob/master/docs/api/modules/data.rst#api-reference

src/gluonnlp/data/transforms.py
```diff
@@ -742,6 +757,7 @@ def __call__(self, sample):
 
     def _tokenize(self, text):
         """Tokenizes a piece of text."""
+        text = convert_to_unicode(text)
```
eric-haibin-lin: In https://github.com/dmlc/gluon-nlp/pull/464#issuecomment-449557024 we decide to not perform unicode conversion inside the tokenizer. The user needs to make sure the inputs are unicode (or he/she can call UnicodeTransform to convert all inputs). Please do not add it here

src/gluonnlp/data/transforms.py
```diff
@@ -988,10 +1004,261 @@ def convert_tokens_to_ids(self, tokens):
         return self.vocab.to_indices(tokens)
 
 
+class BERTTransform(object):
+    """BERT style data transformation.
+
+    Parameters
```
eric-haibin-lin: As we move code to API we need more unit tests to make sure it is not broken in the future. Do you mind adding one?

src/gluonnlp/data/mrpc.py
```diff
@@ -0,0 +1,51 @@
...
+
+@register(segment=['train', 'dev', 'test'])
+class MRPCDataset(TSVDataset):
+    """The Microsoft Research Paraphrase Corpus dataset.
+
+    Parameters
+    ----------
+    segment : str or list of str, default 'train'
+        Dataset segment. Options are 'train', 'val', 'test' or their combinations.
+    root : str, default '$GLUE_DIR/MRPC'
```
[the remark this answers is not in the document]
kenjewu: I checked the license for it, we really can't host it, I will move it back to scripts/bert later.

src/gluonnlp/data/transforms.py
kenjewu: Ok, I am converting a BERTVocab json file and uploading it. I will modify this once the #509 test passed and is merged.

src/gluonnlp/data/transforms.py
kenjewu: ok.

src/gluonnlp/data/transforms.py
kenjewu: Ok, this is easy to modify.

src/gluonnlp/data/transforms.py
kenjewu: Yep. I now get the discussion in #464, and I will remove it.

src/gluonnlp/data/transforms.py
kenjewu: Yes, this is supposed to be. What I want is to add a unit test after the API code meets the requirements.

src/gluonnlp/data/transforms.py
```diff
@@ -988,10 +1004,261 @@ def convert_tokens_to_ids(self, tokens):
...
 def _whitespace_tokenize(text):
     """Runs basic whitespace cleaning and splitting on a piece of text."""
     text = text.strip()
     if not text:
         return []
     tokens = text.split()
     return tokens
+
+
+def convert_to_unicode(text):
```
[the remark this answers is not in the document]
kenjewu: I think this is fine. With a ` UnicodeTransfrom `, the user can explicitly convert related data.  It is no longer necessary to convert data to unicode in the tokenizer.

src/gluonnlp/data/transforms.py
```diff
@@ -22,16 +22,19 @@
...
            'NLTKStanfordSegmenter', 'SentencepieceTokenizer',
-           'SentencepieceDetokenizer', 'BasicTokenizer', 'BERTTokenizer']
+           'SentencepieceDetokenizer', 'BasicTokenizer', 'BERTTokenizer', 'convert_to_unicode',
+           'BERTTransform', 'BERTClassificationTransform']
 
 import os
 import warnings
 
 import unicodedata
+import six
```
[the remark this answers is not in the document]
kenjewu: Ok, if we don't use unicode conversion in the tokenizer or add an ` UnicodeTransform ` as described in the comments below,  then ` six ` may not be used here.

src/gluonnlp/data/transforms.py
kenjewu: I will add this after the API has met the requirements.

src/gluonnlp/data/transforms.py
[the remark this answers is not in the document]
leezu: Why do we need to handle conversion to Unicode at all? Usually the file reading (CorpusDataset etc) are reading Unicode data already.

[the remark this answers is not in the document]
leezu: Also the current implementation assumes that the text is already utf-8, which is a unciode encoding. In this case, data should have already been read to Unicode string. So I'm not sure what the aim is here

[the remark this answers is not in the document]
leezu: Also the current implementation assumes that the text is already utf-8, which is a unciode encoding. In this case, data should have already been read to Unicode string. So I'm not sure what the aim is here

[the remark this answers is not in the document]
eric-haibin-lin: @leezu thanks for the elaboration! Good that our Dataset API already handles encoding. As a more general question - is there a general guideline/assumption that gluonnlp code should always work with unicode? I do see lots of `open` calls in gluonnlp without explicitly specifying encoding. Do we want to recommend `io.open/codes.open` with encoding argument in future code reviews? 

On the other hand, `str` type stand for different things in py2 and py3. Maybe we should make some note what `str` stands for in general in gluonnlp APIs, as we don't want to write "unicode for py2, str in py3" for these arguments... 

@kenjewu at this point I think we should remove `convert_to_unicode ` function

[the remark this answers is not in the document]
leezu: @eric-haibin-lin I agree with all your points

tests/unittest/test_transforms.py
```diff
@@ -212,8 +217,93 @@ def test_berttokenizer():
...
+    assert all(token_ids == valid_token_ids)
+    assert length == len(vocab_tokens) + 3
+    assert all(type_ids == valid_type_ids)
+    assert all(label_ids == np.array([label_cls], dtype=np.int32))
+
+    # test BERTDatasetTransform for regression task
+    label_reg = 0.2
+    bert_reg_dataset_t = t.BERTDatasetTransform(tokenizer, [label_reg], 15, pad=True,
+                                                pair=True, label_dtype='float32')
+    token_ids, length, type_ids, label_value = bert_cls_dataset_t((text_a, text_b, label_cls))
```
eric-haibin-lin: Please test with label_reg instead

src/gluonnlp/data/transforms.py
```diff
@@ -988,10 +1009,246 @@ def convert_tokens_to_ids(self, tokens):
...
+            total_length = len(tokens_a) + len(tokens_b)
+            if total_length <= max_length:
+                break
+            if len(tokens_a) > len(tokens_b):
+                tokens_a.pop()
+            else:
+                tokens_b.pop()
+
+
+class BERTDatasetTransform(object):
```
eric-haibin-lin: On second thought this is very task specific transformation code, compared to `BERTSentenceTransform `. `BERTSentenceTransform` can be generally used as a building block for many tasks like regression, classification and QA. Would it be better to put this transform back to the /script/bert folder? Here's my thought:

Instead, we should need to build composable transformation pipelines:

Currently the `Dataset` API supports `dataset.transform(t1).transform(t2) ... `, where both `t1` and `t2` transform all fields in the input. However, sometimes we want to reuse some `Transform` API for a subset of the fields in the input. For example, say we have a `Dataset` of 3 fields: text_a, text_b and label. A desirable API should allow:
```
trans_fn = ComposableTransform([Tokenizer(), Tokenizer(), LabelTransform()])
dataset.transform(trans_fn)
```
here the 3 fields in a data sample are transformed with 3 transforms respectively. If this is reasonable we can consider doing this in future PRs

@kenjewu @leezu @szha what do you guys think?

szha: batchify has an interface for a similar purpose, called [Tuple](http://gluon-nlp.mxnet.io/master/api/modules/data.batchify.html#gluonnlp.data.batchify.Tuple).

szha: It's probably a good idea to have the same for transform.

szha: Note that batchify function deals with a different signature, so for simple sample-wise transforms, the tuple function may need to be different.

kenjewu: Ok, thank you very much for your comments.

eric-haibin-lin:
@haven-jeon FYI your PR may be affected.

src/gluonnlp/data/transforms.py
```diff
@@ -988,10 +1010,153 @@ def convert_tokens_to_ids(self, tokens):
...
+        vocab = self._tokenizer.vocab
+        tokens = []
+        segment_ids = []
+        tokens.append(vocab.cls_token)
+        segment_ids.append(0)
+        for token in tokens_a:
+            tokens.append(token)
+            segment_ids.append(0)
+        tokens.append(vocab.sep_token)
+        segment_ids.append(0)
```
szha: seems that `segment_ids = [0] * len(tokens)` here would do the job.

src/gluonnlp/data/transforms.py
```diff
@@ -988,10 +1010,153 @@ def convert_tokens_to_ids(self, tokens):
...
+
+        if tokens_b:
+            for token in tokens_b:
+                tokens.append(token)
+                segment_ids.append(1)
+            tokens.append(vocab.sep_token)
+            segment_ids.append(1)
```
szha: here you can do `segment_ids += [1] * (len(tokens) - len(segment_ids)`

scripts/bert/utils.py
```diff
@@ -102,3 +103,40 @@ def read_tf_checkpoint(path):
...
+            token = token.strip()
+            vocab[token] = index
+            index += 1
+    return vocab
+
+
+def printable_text(text):
+    """Returns text encoded in a way suitable for print."""
+    # These functions want `str` for both Python2 and Python3, but in one case
+    # it's a Unicode string and in the other it's a byte string.
```
leezu: Why do we have byte-string `str` in Py2 and not unicode-string `u`? Maybe the data reading code can be changed somewhere to read unicode both on Py2 and Py3. We should try to be consistent and use either byte on both Py2 and Py3, or unicode on both. Less chances for unintended bugs then.

In summary: Is this function really required? Or does the need for this function just point out that some other code does not handle the data reading correctly?

src/gluonnlp/data/transforms.py
```diff
@@ -17,20 +17,22 @@
 # specific language governing permissions and limitations
 # under the License.
 
-# pylint: disable=invalid-encoded-data
+# pylint: disable=invalid-encoded-data, too-many-lines
```
leezu: How about splitting up this file as it seems to be getting too large? Can also be addressed in a separate PR if you feel uncomfortable doing it here & now

scripts/bert/utils.py
kenjewu: @leezu This function is mainly designed for specific string generation in ` create_pretraining_data.py `.  It has not been considered before that it has been assumed that the user input data is already unicode encoded. If on the assumpaion as discussed earlier,  I might be able to remove the byte consideration in code.  @eric-haibin-lin

src/gluonnlp/data/transforms.py
eric-haibin-lin: What about moving XXTokenizers to tokenizers.py, and leave the remaining ones in transforms.py? I'm also ok with doing that in a separate PR.

leezu: Yes, separate PR is also fine. Besides, haibins suggestion, alternatively creating a folder 'transforms' with an init file and splitting the current file up should be possible

src/gluonnlp/data/transforms.py
```diff
@@ -881,6 +895,14 @@ def _is_whitespace(self, char):
             return True
         return False
 
+    def _whitespace_tokenize(self, text):
+        """Runs basic whitespace cleaning and splitting on a piece of text."""
+        text = text.strip()
+        if not text:
```
leezu: The if is not needed. If the string is empty, an empty list is returned by split(). Or do we need to handle None?

src/gluonnlp/data/transforms.py
```diff
@@ -704,26 +718,26 @@ class BasicTokenizer():
     performs invalid character removal (e.g. control chars) and whitespace.
     tokenize CJK chars.
```
leezu: Related to previous PR, should this really be called 'BasicTokenizer? As far as in understand, splitting CJK is useful for BERT models. But BasicTokenizer somewhat implies this should be used / useful everywhere. What do you think @eric-haibin-lin With respect to English models, replacing a single unknown (Chinese) word, with X unknown (Chinese) characters may have an adverse effect?

mli:
Job PR-526/21 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-526/21/index.html

src/gluonnlp/data/transforms.py
```diff
@@ -704,26 +718,26 @@ class BasicTokenizer():
...
     splits punctuation on a piece of text.
-    strips accents and convert to lower case.(If lower_case is true)
+    strips accents and convert to lower case.(If do_lower_case is true)
 
     Parameters
     ----------
-    lower_case : bool, default True
+    do_lower_case : bool, default True
```
fierceX: Why change `lower_case` to `do_lower_case`，I also think that lower_case is more concise. https://github.com/dmlc/gluon-nlp/pull/464#discussion_r243149189

src/gluonnlp/data/transforms.py
kenjewu: Reasonable hints, maybe ` BERTBasicTokenizer ` is more appropriate.  How is it? @eric-haibin-lin @leezu

src/gluonnlp/data/transforms.py
leezu: Why don't we follow Python standard library and call it consistently `lower`? See https://docs.python.org/3/library/stdtypes.html#str.lower
Btw, lowercase can be both state and action. See eg. https://en.wiktionary.org/wiki/lowercase#English. lowercase is a valid verb.

eric-haibin-lin: "lower" sounds good!

kenjewu: It has been modified

src/gluonnlp/data/transforms.py
kenjewu: Yes, thank you for your reminder, it has been revised.

mli:
Job PR-526/22 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-526/22/index.html

mli:
Job PR-526/23 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-526/23/index.html

Issue #577, opened 2019-01-30
[word2vec] Add word2vec binary load function


imgarylai:
I'm willing to add this feature to gluon-nlp. 

Our lab is using gensim to load word2vec binary file and we want to get rid of it. @szha and I have discussed with issue before. This would be great that gluon-nlp can have an interface to load our own pre-trained fastText binary. 

@szha @leezu, please let me know how can I help you guys for this.

szha:
Hi @imgarylai. Thanks for offering to help! For fastText format, there's a `load_fasttext_format` method in the [FasttextEmbeddingModel](http://gluon-nlp.mxnet.io/api/modules/model.train.html#gluonnlp.model.train.FasttextEmbeddingModel) class that can be used without gensim.

szha:
Alternatively, the word2vec format is still missing. Would that be something you're interested in too?

imgarylai:
Yes. I have changed the title. 😝

leezu:
Thanks @imgarylai 

I guess the best way forward would be to just specialize / overwrite the `from_file` for the `Word2Vec` class https://github.com/dmlc/gluon-nlp/blob/master/src/gluonnlp/embedding/token_embedding.py#L961 

(For fastText this was instead added in the `model.train.embedding` API as the binary format does not contain the "final" vectors but rather the vectors for all words / ngrams that must still be combined. But this does not apply in the W2V case.)

szha:
@imgarylai Word2Vec class was written in the way that requires source so that automatic downloading is easier. You can think of the `from_file`/`load_w2v_binary` method as another constructor independent of the source abstraction.

imgarylai:
Move to #620

PR #620, opened 2019-03-06
[FEATURE] offer load_w2v_binary method to load w2v binary file

imgarylai:
## Description ##

This pull request is to propose the `load_w2v_binary` for loading binary pretrained word2vec file into  `Word2Vec` embedding. 

### Essentials ###
- [x] Code is well-documented

### Changes ###
- [x] add `load_w2v_binary`

## Comments ##
- It has been tested through python console. It can load the pretrained w2v binary ,[GoogleNews-vectors-negative300.bin.gz.](https://drive.google.com/file/d/0B7XkCwpI5KDYNlNUTTlSS21pQmM/edit?usp=sharing), successfully. <del>Need helps to the tests. (e.g. where is the binary file would be placed on CI server?) </del>
- I adapt part of code from [gensim](https://github.com/RaRe-Technologies/gensim). Do we have to include license of it? 
- This function overwrite many variables from `__init__`. This part might need to be refactored. 
- Move #577 to here

## Demo

```
# example
w2v = '/home/ubuntu/data/w2v/GoogleNews-vectors-negative300.bin'
from gluonnlp.embedding.token_embedding import Word2Vec
word2vec = Word2Vec()
w.load_w2v_binary(w2v)
w.token_to_idx['hello']
> 20398
w.idx_to_vec[20398]
> [-5.73730469e-02  5.81054688e-02  1.15966797e-03  4.02343750e-01
 -2.22656250e-01 -2.73437500e-01  3.73535156e-02  5.11718750e-01
 ...
 -2.75390625e-01 -4.54101562e-02  1.20117188e-01  1.11816406e-01]
<NDArray 300 @cpu(0)>
len(w.idx_to_vec[20398])
> 300
```


mli:
Job PR-620/1 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-620/1/index.html

imgarylai:
> Nice! Would you mind adding a test?

Sure, tests were added.

src/gluonnlp/embedding/token_embedding.py
```diff
@@ -999,6 +999,8 @@ class Word2Vec(TokenEmbedding):
     embedding_root : str, default '$MXNET_HOME/embedding'
         The root directory for storing embedding-related files.
         MXNET_HOME defaults to '~/.mxnet'.
+    preload : bool, default True
+        Decide to load default model or not.
```
szha: if the goal is to construct an empty Word2Vec class object, how about supporting `source=None`?

imgarylai: right, let me update it.

mli:
Job PR-620/2 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-620/2/index.html

mli:
Job PR-620/3 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-620/3/index.html

src/gluonnlp/embedding/token_embedding.py
```diff
@@ -995,8 +994,9 @@ class Word2Vec(TokenEmbedding):
 
     Parameters
     ----------
-    source : str, default 'GoogleNews-vectors-negative300'
+    source : str or None, default 'GoogleNews-vectors-negative300'
         The name of the pre-trained token embedding file.
+        If source is None, construct an empty Word2Vec class object.
```
leezu: The initializer should actually initialize the class. It seems the currently intended behavior is that users initialize with `source=None` and then call `load_w2v_binary`. How about instead renaming `load_w2v_binary` to something like `from_binary` and making it a constructor too? (ie. need to decorate it with `@classmethod`)

src/gluonnlp/embedding/token_embedding.py
```diff
@@ -1020,9 +1020,85 @@ class Word2Vec(TokenEmbedding):
...
-        self._load_embedding(pretrained_file_path, elem_delim=' ')
+    def load_w2v_binary(self, pretrained_file_path, encoding='utf8'):
+        """Load embedding vectors from a binary pre-trained token embedding file.
+
+        Parameters
+        ----------
+        pretrained_file_path: str
+            The path to a binary pre-trained token embedding file.
+        encoding: str
+            The encoding type of the file.
```
leezu: Why does the binary format have a encoding parameter?

imgarylai: It helps to read word2vec from other languages, e.g. Chinese. For example, all Chinese chars/tokens can be used in unicode(utf8) by passing encoding param. (Actually, utf-8 is the default value already)

src/gluonnlp/embedding/token_embedding.py
imgarylai: Thanks for your review. Let me work on it!

mli:
Job PR-620/1 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-620/1/index.html

src/gluonnlp/embedding/token_embedding.py
```diff
@@ -1020,9 +1020,100 @@ class Word2Vec(TokenEmbedding):
...
+        """
+        embedding = cls(None)
```
imgarylai: I want to discuss about this line. Should I use `embedding = TokenEmbedding(**kwargs)` or current implementation.

mli:
Job PR-620/4 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-620/4/index.html

src/gluonnlp/embedding/token_embedding.py
szha: I think this should probably be named as from_w2v_binary, as the binary format is different from fasttext binary. Once it has this naming, then returning Word2Vec embedding object makes sense.

imgarylai: renamed in a4e5f6a

mli:
Job PR-620/6 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-620/6/index.html

src/gluonnlp/embedding/token_embedding.py
leezu: The current implementation requires that creation of an empty `Word2Vec` object is possible, for which there is actually no use besides the private `_load_w2v_binary`. Switching to `embedding = TokenEmbedding(**kwargs)` results in initializing / returning a parent class object instead of `Word2Vec`. Both seems ok, but not ideal. I'm ok to keep it. Alternatively, taking the `source` argument of `__init__` to mean either the name (current functionality) or path to binary could be possible?

My suggestion for a separate `from_binary` initializer was based on the `model.train.embedding` https://github.com/dmlc/gluon-nlp/blob/d49a789/src/gluonnlp/model/train/embedding.py#L232. In that case, there was no `source` argument whose meaning could be overloaded, and introducing one wouldn't make sense either as it would conflict with other initializer arguments. Here the case is different. Sorry for not pointing this out before.

What do you think?

imgarylai: @leezu thanks for your review. I agree with you either using overload or parent object is not ideal. However, I might not fully understand what you really want to implement it. I will try to explain what I have thoughts. 

Do you mean the constructor can be implemented like this? However, the error would be unclear for users, right?
```py
class Word2Vec(TokenEmbedding):
    source_file_hash = C.WORD2VEC_NPZ_SHA1
    def __init__(self, source='GoogleNews-vectors-negative300',
                 embedding_root=os.path.join(get_home_dir(), 'embedding'), encoding='utf8',
                 **kwargs):
        super(Word2Vec, self).__init__(**kwargs)
        try:
            self._check_source(self.source_file_hash, source)
            pretrained_file_path = self._get_file_path(self.source_file_hash,
                                                       embedding_root, source)
            self._load_embedding(pretrained_file_path, elem_delim=' ')
        except KeyError:
            pretrained_file_path = os.path.expanduser(source)
            if is_binary(pretrained_file_path): 
                self._load_w2v_binary(pretrained_file_path, encoding=encoding)
            raise TypeError('{} is not a binary file'.format(pretrained_file_path))
```

leezu: Yes, I meant such `__init__` as in your example. Do you think it is more appropriate than introducing a second constructor?

To clarify the error message: If source neither refers to a valid name for `self._load_embedding` nor to a valid path (ie. `except KeyError`), raise a generic error that `source` must be either a valid name or a path to binary file.

Only `raise TypeError('{} is not a binary file'.format(pretrained_file_path))` if `source` is not a valid name, but a valid path to an existing file, which however is not of the correct binary format.

What do you think?

imgarylai: @leezu Thanks. 
The code has been updated, but this implementation doesn't have the `is_binary`. I decide to remove it because there is no a really good way to check a binary file in python. And I think if the file is not in a good format, errors will be raised in `_load_w2v_binary`. 
For text file, it is loaded from `from_file`, so it doesn't go through this constructor.  

Does it make sense to you?

szha: Let's stick to a convention for binary files. I think an extension `.bin` is good enough for representing it for now.

mli:
Job PR-620/13 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-620/13/index.html

src/gluonnlp/embedding/token_embedding.py
```diff
@@ -1019,10 +1019,103 @@ class Word2Vec(TokenEmbedding):
...
+        self._idx_to_vec = nd.array(all_elems).reshape((-1, vec_len))
+
+        if self.unknown_token:
+            if loaded_unknown_vec is None:
+                self._idx_to_vec[C.UNK_IDX] = self._init_unknown_vec(shape=vec_len)
+            else:
+                self._idx_to_vec[C.UNK_IDX] = nd.array(loaded_unknown_vec)
+
+    @classmethod
+    def from_w2v_binary(cls, pretrained_file_path, encoding='utf8'):
```
leezu: I think this can be deleted, given that `__init__` now handles loading binary files? If so, the doc of `__init__` still needs to be updated

imgarylai: Keeping this might not be a bad idea. `from_w2v_binary` is align with `from_file`. Both load pre-trained model outside from the source list. Despite that the `__init__` constructor handle binary files, having this function has a similar user interface for user to load pretrained models outside from our source list. 
How do you think? 

Thanks for your review. I updated the doc as well.

mli:
Job PR-620/16 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-620/16/index.html

src/gluonnlp/embedding/token_embedding.py
leezu: Thanks @imgarylai! I'm fine with keeping it. While it duplicates functionality indeed it may be easier to discover the functionality for some users.

In that case, can you update the doc to clarify that this is equivalent to `__init__`? Otherwise users may wonder about the difference. Regarding the doc of `__init__`, 

> The name of the pre-trained token embedding file.

is not very clear. Can you clarify that either the name of one of the included embeddings, or a path to a binary file can be provided? Thank you!

imgarylai: @leezu Yes, sure!
The doc has been updated.

mli:
Job PR-620/20 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-620/20/index.html

eric-haibin-lin:
@imgarylai thanks for the contribution!

PR #624, opened 2019-03-08
[FEATURE] Add static bert

cgraywang:
## Description ##

This PR produces a standalone script repo for hybridizable BERT of static shape, with an example of using the hybridized BERT for QA, where a more general usage guideline is also summarized.

### Essentials ###
- [X] PR's title starts with a category (e.g. [BUGFIX], [MODEL], [TUTORIAL], [FEATURE], [DOC], etc)
- [x] Code is well-documented

### Changes ###
- Add hybridizable BERT of static shape
- Add an example of using the hybridized BERT for QA
- Add a more general usage guideline

## Comments ##


scripts/bert/index.rst
```diff
@@ -279,3 +279,16 @@ Command line interface
        -0.42823148, -0.48873612], dtype=float32), array([-0.6576557 , -0.09822252,  0.1121515 , ..., -0.21743725,
        -0.1820574 , -0.16115054], dtype=float32)]
 
+.. container:: cards
+
+   .. card::
+      :title: Static BERT
+      :link: staticbert/index.html
+
+      Hybridizable BERT of static shape with an example on SQuAD.
```
szha: remove

szha: also I think this is a feature for exporting bert model so it probably shouldn't be in a separate page.

eric-haibin-lin:
Agree, can we add export option to existing finetune classification/qa scripts?

scripts/bert/staticbert/bert_qa_dataset.py
```diff
@@ -0,0 +1,506 @@
+# coding=utf-8
```
eric-haibin-lin: can we reuse ../qa_dataset.py?

scripts/bert/staticbert/bert_qa_evaluate.py
```diff
@@ -0,0 +1,409 @@
+# coding=utf-8
```
eric-haibin-lin: can we reuse ../qa_evaluate.py? maybe a symbolic link just works

szha:
Users may not realize what staticbert means and why it should be static (i.e. implementation detail). In terms of functionality, this is purely for the purpose exporting model (i.e. actual use case). Thus, the arrangement of the script should reflect the use case, not the implementation detail.

scripts/bert/staticbert/index.rst
```diff
@@ -0,0 +1,166 @@
+Example Usage of Exporting Hybridizable BERT
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+For exporting hybridizable BERT, we mean the BERT with fixed input embedding size and sequence length can be exported through
```
eric-haibin-lin: Can we add two more sentences for the context? 

The BERTModel class is a subclass of Block, rather than HybridBlock. To support exporting BERT model to json format for deployment, we introduce the StaticBERT class.  Specifically, by exporting hybridizable BERT, we mean the BERT model with ...

scripts/bert/staticbert/index.rst
```diff
@@ -0,0 +1,166 @@
...
+An example can be found in 'static_bert_for_qa_model.py'.
+
+
+Step 3: specify the static input size and sequence length of the input data via environment variables
+in the training script:
+
+.. code-block:: console
+
+    $ os.environ['SEQLENGTH'] = str(args.seq_length)
+    $ os.environ['INPUTSIZE'] = str(args.input_size)
```
eric-haibin-lin: SEQLENGTH is easy to understand. What does inputsize refer to?

scripts/bert/staticbert/index.rst
```diff
@@ -0,0 +1,166 @@
...
+
+
+Step 5: export trained model:
+
+.. code-block:: console
+
+    $ net.export(os.path.join(args.output_dir, 'static_net'), epoch=args.epochs)
+
+To export the model, in 'static_export_squad.py', set export=True.
+
```
eric-haibin-lin: Can we add an example usage like the following?
```
python static_export_squad.py --export --model_parameters /PATH/TO/PARAMS/FROM/BERTQA/BLOCK 
```
and explain that this loads the the StaticBERTQA HybridBlock with parameter trained by a normal BERTQA Block, and export the HybridBlock to json format.

scripts/bert/staticbert/static_finetune_squad.py
```diff
@@ -0,0 +1,533 @@
...
+                log_num = 0
+        epoch_toc = time.time()
+        log.info('Epoch: {}, Time cost={:.2f} s, Thoughput={:.2f} samples/s'
+                 .format(epoch_id, epoch_toc - epoch_tic,
+                         len(train_dataloader) / (epoch_toc - epoch_tic)))
+
+    net.save_parameters(os.path.join(output_dir, 'net.params'))
+
+
+def evaluate():
```
szha: Evaluating the HybridBlock version of the BERT model does not add much value in educating users, especially given that this is already avaiable in the other static_bert_qa script. If you think showing the evaluation is useful, consider making the evaluation optional, and use symbol block with the exported symbol and parameters instead.

scripts/bert/staticbert/index.rst
```diff
@@ -0,0 +1,166 @@
+Example Usage of Exporting Hybridizable BERT
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+For exporting hybridizable BERT, we mean the BERT with fixed input embedding size and sequence length can be exported through
+a static shape based implementation of hybridblock based BERT. By using this, we can export a block based BERT model.
+
+Please follow the steps below for exporting the model.
+
+Step 1: Load dataset. An example can be found in 'static_export_squad.py'.
```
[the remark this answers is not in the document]
eric-haibin-lin: What about using a dummy data batch?

scripts/bert/staticbert/static_export_squad.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+###############################################################################
+#        Specify the static input size and sequence length                    #
+###############################################################################
+
+#SEQLENGTH stands for the sequence length of the input,
+#INPUTSIZE represents the embedding size of the input.
+os.environ['SEQLENGTH'] = str(args.seq_length)
+os.environ['INPUTSIZE'] = str(args.input_size)
```
szha: use command line options instead of environment variables. there's no need for more than one way of such input

scripts/bert/staticbert/static_export_squad.py
```diff
@@ -0,0 +1,245 @@
...
+doc_stride = args.doc_stride
+max_query_length = args.max_query_length
+
+if max_seq_length <= max_query_length + 3:
+    raise ValueError('The max_seq_length (%d) must be greater than max_query_length '
+                     '(%d) + 3' % (max_seq_length, max_query_length))
+
+
+###############################################################################
+#                              Load datasets                                  #
```
szha: this would be more accurately described as "preparing dummy input data"

scripts/bert/staticbert/static_export_squad.py
cgraywang: This is used in `StaticBaseTransformerEncoder`, command line options wouldn't work.

szha: I'm not sure I follow. IMO the class should take these as additional arguments.

mli:
Job PR-624/5 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-624/5/index.html

scripts/bert/staticbert/static_export_squad.py
cgraywang: As I said, please take a look at `StaticBaseTransformerEncoder`, we are doing this 
`        length = int(os.environ['SEQLENGTH'])
        C_in = int(os.environ['INPUTSIZE'])` in the `hybrid_forward`. This is how static hybridizable BERT works. Otherwise, we need to modify the `StaticBERTModel` structure, then `get_bert_model` wouldn't work since it has different model architecture compared to pre-defined `BERTModel`.

szha: Thanks for the explanation. I suggest that we pass these arguments from the constructor of that block and update the static-BERT-specific `get_bert_model` method accordingly. Passing argument through an environment variable seems too hacky for the general users.

cgraywang: Given that this is in the script and not in API, and the `export static bert` is an experimental API, and the current functionality doesn't affect the users. I would like to have this one for now. Besides, since I am not sure whether changing the model structure might affect parameter loading (when pretrained=True) in `get_static_bert_model` (this is from `BERTModel`), and this will generate more reductant codes such as pretrained settings, and might possibly involves more tests.

szha: > the current functionality doesn't affect the users

The scripts are to be used by regular users who want to try to deploy this model, so this is not true. Generally we don't check in code that won't be used.

> I am not sure whether changing the model structure might affect parameter loading

As long as you load the length from the parameter saved in the constructor in the places where it's reading from environment variables, the model structure is not really changed so it doesn't affect pre-trained parameter loading. Feel free to let me know if you need help on this.

> more reductant codes such as pretrained settings

Copying the pretrained settings shouldn't be necessary. They are just dictionaries that can be modified in the methods, and you can easily add new entries by `pretrained_parameters['seq_length'] = X`. It shouldn't require duplicating the parameters.

If you run into trouble feel free to ping me and I'm happy to offer help.

szha: Having a consistent design of the interface goes a long way in ensuring that things remain extensible and useful. The code that we write once will potentially be used and read by many people, for many times. Hope that you understand.

cgraywang: >> the current functionality doesn't affect the users

>The scripts are to be used by regular users who want to try to deploy this model, so this is not true. Generally we don't check in code that won't be used.

Users won't need to write any code but just specify the `input_size` and `seq_length` in the command line, the usage of the `input_size` and `seq_length` in hiding in the API. It doesn't affect the users. 

>Having a consistent design of the interface goes a long way in ensuring that things remain extensible and useful. The code that we write once will potentially be used and read by many people, for many times. Hope that you understand.

I agree, that's the reason we have it now in the SCRIPT not API as an experimental API. If we would like to spend more time on this, I am also fine with it, just in terms of time, I need to spend more time on testing it out. Hope that you can understand that as well.

szha: I think you can skip rerunning the squad experiment. This is not a requirement of the export script.

mli:
Job PR-624/6 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-624/6/index.html

scripts/bert/index.rst
```diff
@@ -279,3 +279,168 @@ Command line interface
...
+a static shape based implementation of hybridblock based BERT. By using this, we can export a block based BERT model.
+
+Please follow the steps below for exporting the model.
+
+
+Step 1: create a hybridizable task guided model using BERT:
+
+.. code-block:: console
+
+    $ class StaticBertForQA(HybridBlock)
```
szha: This should be python code block without the dollar prefix.

scripts/bert/index.rst
```diff
@@ -279,3 +279,168 @@ Command line interface
...
+
+An example can be found in 'staticbert/static_bert_for_qa_model.py'.
+
+
+Step 2: hybridize the model in the script:
+
+.. code-block:: console
+
+    $ net = StaticBertForQA(bert=bert)
+    $ net.hybridize(static_alloc=True, static_shape=True)
```
szha: same here

scripts/bert/index.rst
```diff
@@ -279,3 +279,168 @@ Command line interface
...
+
+An example can be found in 'staticbert/static_export_squad.py'.
+
+
+Step 3: export trained model:
+
+.. code-block:: console
+
+    $ net.export(os.path.join(args.output_dir, 'static_net'), epoch=args.epochs)
```
szha: same here

scripts/bert/index.rst
```diff
@@ -279,3 +279,168 @@ Command line interface
...
+
+This example mainly introduces the steps needed to use the hybridizable BERT models to finetune on a specific NLP task.
+We use SQuAD dataset for Question Answering as an example.
+
+
+Step 1: create a hybridizable task guided model using BERT:
+
+.. code-block:: console
+
+    $ class StaticBertForQA(HybridBlock)
```
szha: same here. also, some contents here duplicate with the above section. consider referring to the previous section instead.

scripts/bert/index.rst
```diff
@@ -279,3 +279,168 @@ Command line interface
...
+
+An example can be found in 'staticbert/static_bert_for_qa_model.py'.
+
+
+Step 2: hybridize the model in the training script:
+
+.. code-block:: console
+
+    $ net = StaticBertForQA(bert=bert)
+    $ net.hybridize(static_alloc=True, static_shape=True)
```
szha: same, python block

scripts/bert/index.rst
```diff
@@ -279,3 +279,168 @@ Command line interface
...
+
+An example can be found in 'staticbert/static_finetune_squad.py'.
+
+
+Step 3: export trained model:
+
+.. code-block:: console
+
+    $ net.export(os.path.join(args.output_dir, 'static_net'), epoch=args.epochs)
```
szha: same

cgraywang:
It looks like there could be an issue in the last CI run, so @eric-haibin-lin helps to re-trigger the CI.

mli:
Job PR-624/10 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-624/10/index.html

mli:
Job PR-624/9 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-624/9/index.html

TaoLv:
Hi @cgraywang what's the difference in between the saved net.params and static_net-0002.params?

cgraywang:
@TaoLv `net.params` can only be loaded using `load_parameters `, since it doesn't have the model structure. If you also want to save the model structure, `HybridBlock.export()` would do that.

PR #648, opened 2019-03-29
[BUGFIX] Fix _get_rnn_cell

MarisaKirisame:


mli:
Job PR-648/1 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-648/1/index.html

src/gluonnlp/model/utils.py
```diff
@@ -199,11 +199,11 @@ def _get_rnn_cell(mode, num_layers, input_size, hidden_size,
         Only available when the mode=lstmpc.
     """
 
-    assert mode == 'lstmpc' and proj_size is not None, \
+    assert mode == 'lstmpc' or proj_size is None, \
```
cgraywang: Nice catch! `proj_size` is required for `lstmpc`. One simple modification would be removing `mode == lstmpc`, and add a condition `if mode == 'lstmpc'`.

eric-haibin-lin: Sorry in the last review I didn't realize this change will break lstmpc cell. What about moving the assertion inside `LSTMPCellWithClip ` to check `proj_size`, `cell_clip ` and `proj_clip `? https://github.com/dmlc/gluon-nlp/blob/aa2c7e6a7bcf83ed73b6e0f46ee5b48f15ee82c4/src/gluonnlp/model/lstmpcellwithclip.py#L85-L97

szha:
Thanks for the fix, @MarisaKirisame. Would you have some time to address the comments? Let me know how I can help.

MarisaKirisame:
@szha hi, sorry for the delay! I was pushing for a conference deadline for a week (this come up). Now that it is over I can actually write other code. I will look at this and fix ASAP.

mli:
Job PR-648/3 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-648/3/index.html

mli:
Job PR-648/2 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-648/2/index.html

MarisaKirisame:
@eric-haibin-lin I cannot move them inside.
The problem is, I have to check that if lstmpc is on, it is not None, and if it is off it is None.
The latter must be done outside of LSTMPCellWithClip.
IMO the principle way to fix this is to let _get_rnn_cell take a cell instead of take a mode.
What do you think?

szha:
@eric-haibin-lin ping.

eric-haibin-lin:
@MarisaKirisame sorry for the late reply. If a cell is passed, we still need to contruct a cell somewhere. What about just fixing the assertion in the following way?
```
    if mode != 'lstmpc':
        assert proj_size is None, 'proj_size takes effect only when mode is lstmpc'
        assert cell_clip is None, 'cell_clip takes effect only when mode is lstmpc'
        assert proj_clip is None, 'proj_clip takes effect only when mode is lstmpc'

```

MarisaKirisame:
@eric-haibin-lin that is what the current fix do, it is just in different form. Is there reason to prefer one over the other?

eric-haibin-lin:
nvm. The current fix looks good. Just triggered the CI again. Thanks for the fix!

mli:
Job PR-648/2 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-648/2/index.html

PR #665, opened 2019-04-17
[Enhancement] multi node BERT pre-training with horovod

eric-haibin-lin:
## Description ##
This PR introduces `run_pretraining_hvd.py` which uses horovd for multi-gpu/multi-host training. With synthetic data (bs=24, seq_len=512), the training speedup is 58x using 64 GPUs (p3dn.24xlarge instances).

This PR also 
- refactors the exiting `run_pretraining.py` script. 
- added more script tests

### Essentials ###
- [ ] Code is well-documented

### Changes ###

## Comments ##


mli:
Job PR-665/5 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-665/5/index.html

scripts/bert/fp16_utils.py
```diff
@@ -230,7 +232,7 @@ def __init__(self, init_scale=2.**15, scale_factor=2., scale_window=2000,
         self._last_overflow_iter = -1
         self._last_rescale_iter = -1
         self._overflows_since_rescale = 0
-        self._verbose = verbose
+        self._logger = logging.info if verbose else logging.debug
```
szha: debug log is usually more verbose than info. is this intended?

mli:
Job PR-665/6 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-665/6/index.html

scripts/bert/pretraining_utils.py
```diff
@@ -0,0 +1,288 @@
...
+from mxnet.gluon.data import DataLoader
+
+import gluonnlp as nlp
+from gluonnlp.data.batchify import Tuple, Stack, Pad
+from gluonnlp.metric import MaskedAccuracy
+
+__all__ = ['get_model', 'get_pretrain_dataset', 'get_dummy_dataloader',
+           'save_params', 'evaluate', 'forward', 'split_and_load', 'get_argparser']
+
+def get_model(ctx, model, pretrained, dataset_name, dtype, ckpt_dir=None, start_step=None):
```
szha: it would be confusing to have another version of get model

mli:
Job PR-665/8 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-665/8/index.html

mli:
Job PR-665/9 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-665/9/index.html

eric-haibin-lin:
Follow up tasks: #673 #674 #675

scripts/bert/pretraining_utils.py
```diff
@@ -0,0 +1,288 @@
...
+    stream = nlp.data.SimpleDatasetStream(nlp.data.NumpyDataset, data, split_sampler)
+    if prefetch:
+        stream = nlp.data.PrefetchingStream(stream)
+
+    def get_dataloader(dataset):
+        """create data loader based on the dataset chunk"""
+        lengths = dataset.get_field('valid_lengths')
+        # A batch includes: input_id, masked_id, masked_position, masked_weight,
+        #                   next_sentence_label, segment_id, valid_length
+        batchify_fn = Tuple(Pad(), Pad(), Pad(), Pad(), Stack(), Pad(), Stack())
```
kenjewu: The default pad_val of `Pad()` is 0, but is the pad_val of each field 0?

scripts/bert/pretraining_utils.py
```diff
@@ -0,0 +1,288 @@
...
+    # pylint: disable=line-too-long
+    logging.info('[step {}]\tmlm_loss={:.5f}\tmlm_acc={:.5f}\tnsp_loss={:.5f}\tnsp_acc={:.3f}\tthroughput={:.1f}K tks/s\tlr={:.7f} time={:.2f}, latency={:.1f} ms/batch'
+                 .format(step_num, running_mlm_loss.asscalar(), mlm_metric.get()[1] * 100, running_nsp_loss.asscalar(),
+                         nsp_metric.get()[1] * 100, throughput.asscalar(), lr, duration, duration*1000/log_interval))
+    # pylint: enable=line-too-long
+
+def split_and_load(arrs, ctx):
+    """split and load arrays to a list of contexts"""
+    assert isinstance(arrs, (list, tuple))
+    if len(ctx) == 1:
```
kenjewu: Is it necessary to make this judgment?  This special judgment has been made in the implementation of [`mx.gluon.utils.split_and_load()`](https://mxnet.apache.org/_modules/mxnet/gluon/utils.html#split_and_load).

scripts/bert/pretraining_utils.py
```diff
@@ -0,0 +1,288 @@
...
+                 decoded, masked_weight, ls1, ls2, valid_length) = out
+                loss_list.append(ls)
+                ns_label_list.append(next_sentence_label)
+                ns_pred_list.append(classified)
+                mask_label_list.append(masked_id)
+                mask_pred_list.append(decoded)
+                mask_weight_list.append(masked_weight)
+
+                running_mlm_loss += ls1.as_in_context(mx.cpu())
+                running_nsp_loss += ls2.as_in_context(mx.cpu())
```
kenjewu: Since loss will be copied back to the cpu and the value is scalar, is it better to use `asscalar()` directly instead of in the `log()` function?

scripts/bert/run_pretraining_hvd.py
```diff
@@ -0,0 +1,235 @@
...
+                    nsp_metric.reset_local()
+
+                # saving checkpoints
+                if args.ckpt_dir and (step_num + 1) % (args.ckpt_interval) == 0 \
+                   and (batch_num + 1) % accumulate == 0 and local_rank == 0:
+                    save_params(step_num, model, trainer, args.ckpt_dir)
+
+                batch_num += 1
+
+    if local_rank == 0:
```
kenjewu: Save checkpoints only on worker 0 to prevent other workers from corrupting them. if run a job on 2 machines, and each machine has 4 GPUs, the rank will be a number [0, 7], and the local rank will be a number [0, 3]. So it should be saved when rank is 0.
```python
if args.ckpt_dir and rank == 0
```

scripts/bert/run_pretraining_hvd.py
```diff
@@ -0,0 +1,235 @@
...
+    hvd.broadcast_parameters(model.collect_params(), root_rank=0)
+
+    mlm_metric = nlp.metric.MaskedAccuracy()
+    nsp_metric = nlp.metric.MaskedAccuracy()
+    mlm_metric.reset()
+    nsp_metric.reset()
+
+    logging.debug('Creating distributed trainer...')
+    lr = args.lr
+    optim_params = {'learning_rate': lr, 'epsilon': 1e-6, 'wd': 0.01}
```
kenjewu: Does need to scale the learning rate by number of workers?
```python
lr = lr * num_workers
```

scripts/bert/pretraining_utils.py
eric-haibin-lin: I want to avoid using asscalar() here since it is a blocking call. We want to avoid making these blocking calls so that metric computation and sgd update can be scheduled earlier

scripts/bert/run_pretraining_hvd.py
eric-haibin-lin: I wanted to make sure all machines have the checkpoint, as the checkpoint has to be present at each machine when loading it. There will be 2 processes with local_rank = 0, and only 1 process with rank = 0.

scripts/bert/run_pretraining_hvd.py
eric-haibin-lin: There are studies arguing linear scaling rule vs. sqrt scaling rule vs. fixed value for learning rate with multiple machines. See https://arxiv.org/abs/1808.01371 , "Large-Batch Training for LSTM and Beyond", and https://arxiv.org/abs/1706.02677 
The learning rate is left for the user to decide

scripts/bert/pretraining_utils.py
eric-haibin-lin: Good point. removed

scripts/bert/pretraining_utils.py
eric-haibin-lin: Good catch. I think it's a general issue for gluon-nlp. For this script, the model forward and loss computation relies on the mask_id and valid_length of the data batch. Pad does not affect the correctness. Let me create an issue for this. I think we can print a warning `using default padding value 0` if user does not specify the pad value when using `Pad()`

scripts/bert/run_pretraining_hvd.py
kenjewu: Ok, got it.

mli:
Job PR-665/10 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-665/10/index.html

scripts/bert/run_pretraining_hvd.py
eric-haibin-lin: Also FYI https://github.com/horovod/horovod/blob/master/horovod/mxnet/__init__.py#L96

Issue #727, opened 2019-05-24
pytest 4 support


leezu:
Tests are failing with current pytest 4.5

This does not affect CI as pytest < 4 is specified in the environment, but affects contributors running tests outside of CI.

leezu:
Specifically https://docs.pytest.org/en/latest/historical-notes.html#marker-revamp

leezu:
Fixed by #728

PR #728, opened 2019-05-24
Pytest 4 support

leezu:
Should fix pytest > 4 support. See https://github.com/dmlc/gluon-nlp/issues/727

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -69,6 +69,7 @@ setup(
             'sphinx-gallery',
             'sphinx_rtd_theme',
             'nbsphinx',
+            'flaky',
         ],
     },
 )
--- a/conftest.py
+++ b/conftest.py
@@ -150,7 +150,7 @@ def function_scope_seed(request):
 
     """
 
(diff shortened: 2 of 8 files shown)


setup.py
```diff
@@ -69,6 +69,7 @@ def find_version(*file_paths):
             'sphinx-gallery',
             'sphinx_rtd_theme',
             'nbsphinx',
+            'flaky',
```
szha: flaky doesn't seem used outside of test.

leezu: Yes, it's declared here for 'dev' purpose

leezu: My expectation is that 'pip install .[dev]' will install all development (and testing) dependencies

szha:
Thanks. I was thinking of using tests_require but since we're not testing through setup.py by using test_suite I think it's fine.

mli:
Job PR-728/1 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-728/1/index.html

PR #732, opened 2019-05-27
[FEATURE] Flexible vocabulary

### LINKED ISSUE #393: gluonnlp.Vocab hard coded id of <unk> token
Hi gluon-nlp team,

`gluonnlp.Vocab` always assigns `0` as id of `<unk>`, causing trouble in some situations where non-zero ids represent valid tokens. The same goes to other special tokens. When I need a mapping of `PAD, ROOT, UNK = 0, 1, 2`, the current implementation doesn't allow this. Besides, `<unk>` will be used in decoding phase, so I can't simply disable it.

Is there any possibilities for a new feature enabling flexible assignment of ids to special tokens?

Thanks.

### LINKED ISSUE #572: [Vocab] Support special token registration
per discussion in https://github.com/dmlc/gluon-nlp/issues/492#issuecomment-451769526
We want an extensible way which allows users to register whatever special token fields are needed (except unknown token which has special fallback behavior).

For example:
```
>>> vocab = nlp.Vocab(...)
>>> vocab.register_special_token('mask', '[MASK]')
>>> vocab.mask
'[MASK]'
>>> vocab['mask']
'[MASK]'
```

leezu:
Currently the `gluonnlp.Vocab` always assigns 0 as id of `<unk>` (#393). Further
reserved tokens are not exposed as attributes of the vocab (#572). This PR
addresses both issues.

### Essentials ###
- [X] PR's title starts with a category (e.g. [BUGFIX], [MODEL], [TUTORIAL], [FEATURE], [DOC], etc)
- [X] Code is well-documented

### Changes ###
- [X] Add `token_to_idx` argument to `gluonnlp.Vocab` which allows users to
      optionally specify the index to be associated to a token for a subset of
      tokens in the vocabulary. (Fixes #393)
- [X] Add `identifiers_to_tokens` argument to `gluonnlp.Vocab` which allows users to
      optionally specify a Mapping, specifying identifiers for tokens that will be
      exposed as attributes of the vocabulary (Fixes #572)

## Comments ##
- The `token_to_idx` argument can only contain tokens that will actually be part
  of the vocabulary. On the one hand, a user may not know which tokens will be
  part of the vocabulary, eg. when specifying a maximum size argument. On the
  other hand, if the user does not know if a certain token will be part of the
  vocab or not, it seems unlikely that the user would like to pre-specify the
  index of that token. Thus this shouldn't pose any problems.

### DIFF
--- a/src/gluonnlp/embedding/token_embedding.py
+++ b/src/gluonnlp/embedding/token_embedding.py
@@ -42,6 +42,9 @@ from ..base import get_home_dir
 from ..data.utils import DefaultLookupDict
 from ..model.train import FasttextEmbeddingModel
 
+UNK_IDX = 0   # This should not be changed as long as serialized token
+              # embeddings redistributed on S3 contain an unknown token. Commit
+              # 46eb28ffcf542ab8ed801a727a45404a73adbce3 has more context.
 
 def register(embedding_cls):
     """Registers a new token embedding.
@@ -188,7 +191,7 @@ class TokenEmbedding(object):
         self._unknown_lookup = unknown_lookup
         self._idx_to_token = [unknown_token] if unknown_token else []
         if unknown_token:
-            self._token_to_idx = DefaultLookupDict(C.UNK_IDX)
+            self._token_to_idx = DefaultLookupDict(UNK_IDX)
         else:
             self._token_to_idx = {}
         self._token_to_idx.update((token, idx) for idx, token in enumerate(self._idx_to_token))
@@ -316,9 +319,9 @@ class TokenEmbedding(object):
 
         if self.unknown_token:
             if loaded_unknown_vec is None:
-                self._idx_to_vec[C.UNK_IDX] = self._init_unknown_vec(shape=vec_len)
+                self._idx_to_vec[UNK_IDX] = self._init_unknown_vec(shape=vec_len)
             else:
-                self._idx_to_vec[C.UNK_IDX] = nd.array(loaded_unknown_vec)
+                self._idx_to_vec[UNK_IDX] = nd.array(loaded_unknown_vec)
 
     def _load_embedding_serialized(self, pretrained_file_path):
         """Load embedding vectors from a pre-trained token embedding file.
@@ -334,23 +337,23 @@ class TokenEmbedding(object):
         deserialized_embedding = TokenEmbedding.deserialize(pretrained_file_path)
         if deserialized_embedding.unknown_token:
             # Some .npz files on S3 may contain an unknown token and its
-            # respective embedding. As a workaround, we assume that C.UNK_IDX
+            # respective embedding. As a workaround, we assume that UNK_IDX
             # is the same now as it was when the .npz was generated. Under this
             # assumption we can safely overwrite the respective token and
             # vector from the npz.
             if deserialized_embedding.unknown_token:
                 idx_to_token = deserialized_embedding.idx_to_token
                 idx_to_vec = deserialized_embedding.idx_to_vec
-                idx_to_token[C.UNK_IDX] = self.unknown_token
+                idx_to_token[UNK_IDX] = self.unknown_token
                 if self._init_unknown_vec:
                     vec_len = idx_to_vec.shape[1]
-                    idx_to_vec[C.UNK_IDX] = self._init_unknown_vec(shape=vec_len)
+                    idx_to_vec[UNK_IDX] = self._init_unknown_vec(shape=vec_len)
             else:
                 # If the TokenEmbedding shall not have an unknown token, we
                 # just delete the one in the npz.
-                assert C.UNK_IDX == 0
-                idx_to_token = deserialized_embedding.idx_to_token[C.UNK_IDX + 1:]
-                idx_to_vec = deserialized_embedding.idx_to_vec[C.UNK_IDX + 1:]
+                assert UNK_IDX == 0
+                idx_to_token = deserialized_embedding.idx_to_token[UNK_IDX + 1:]
+                idx_to_vec = deserialized_embedding.idx_to_vec[UNK_IDX + 1:]
         else:
             idx_to_token = deserialized_embedding.idx_to_token
             idx_to_vec = deserialized_embedding.idx_to_vec
@@ -363,10 +366,10 @@ class TokenEmbedding(object):
             try:
                 unknown_token_idx = deserialized_embedding.idx_to_token.index(
                     self.unknown_token)
-                idx_to_token[C.UNK_IDX], idx_to_token[
+                idx_to_token[UNK_IDX], idx_to_token[
                     unknown_token_idx] = idx_to_token[
-                        unknown_token_idx], idx_to_token[C.UNK_IDX]
-                idxs = [C.UNK_IDX, unknown_token_idx]
+                        unknown_token_idx], idx_to_token[UNK_IDX]
+                idxs = [UNK_IDX, unknown_token_idx]
                 idx_to_vec[idxs] = idx_to_vec[idxs[::-1]]
             except ValueError:
                 vec_len = idx_to_vec.shape[1]
@@ -595,7 +598,7 @@ class TokenEmbedding(object):
         if ((self.allow_extend or all(t in self.token_to_idx for t in tokens))
                 and self._idx_to_vec is None):
             # Initialize self._idx_to_vec
-            assert C.UNK_IDX == 0
+            assert UNK_IDX == 0
             self._idx_to_vec = self._init_unknown_vec(
                 shape=(1, new_embedding.shape[-1]))
 
@@ -632,7 +635,7 @@ class TokenEmbedding(object):
                     raise KeyError(('Token "{}" is unknown. To update the embedding vector for an'
                                     ' unknown token, please explicitly include "{}" as the '
                                     '`unknown_token` in `tokens`. This is to avoid unintended '
-                                    'updates.').format(token, self._idx_to_token[C.UNK_IDX]))
+                                    'updates.').format(token, self._idx_to_token[UNK_IDX]))
                 else:
                     raise KeyError(('Token "{}" is unknown. Updating the embedding vector for an '
                                     'unknown token is not allowed because `unknown_token` is not '
@@ -727,7 +730,7 @@ class TokenEmbedding(object):
         if not unknown_token:  # Store empty string instead of None
             unknown_token = ''
         else:
-            assert unknown_token == idx_to_token[C.UNK_IDX]
+            assert unknown_token == idx_to_token[UNK_IDX]
 
         if not compress:
             np.savez(file=file_path, unknown_token=unknown_token,
@@ -773,8 +776,8 @@ class TokenEmbedding(object):
 
         embedding = cls(unknown_token=unknown_token, **kwargs)
         if unknown_token:
-            assert unknown_token == idx_to_token[C.UNK_IDX]
-            embedding._token_to_idx = DefaultLookupDict(C.UNK_IDX)
+            assert unknown_token == idx_to_token[UNK_IDX]
+            embedding._token_to_idx = DefaultLookupDict(UNK_IDX)
         else:
             embedding._token_to_idx = {}
 
@@ -1049,7 +1052,7 @@ class Word2Vec(TokenEmbedding):
         """
         self._idx_to_token = [self.unknown_token] if self.unknown_token else []
         if self.unknown_token:
-            self._token_to_idx = DefaultLookupDict(C.UNK_IDX)
+            self._token_to_idx = DefaultLookupDict(UNK_IDX)
         else:
             self._token_to_idx = {}
         self._token_to_idx.update((token, idx) for idx, token in enumerate(self._idx_to_token))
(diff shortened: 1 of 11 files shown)


mli:
Job PR-732/1 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-732/1/index.html

mli:
Job PR-732/5 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-732/5/index.html

mli:
Job PR-732/6 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-732/6/index.html

eric-haibin-lin:
@rich-junwang  you might be interested in this

rich-junwang:
> @rich-junwang you might be interested in this

Thanks, this is helpful!

leezu:
@rich-junwang feel free to review the code and leave comments

mli:
Job PR-732/7 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-732/7/index.html

leezu:
@kenjewu @eric-haibin-lin please take a look at the refactored BERTVocab.

The only user-visible change is that the static default UNKNOWN_TOKEN, etc attributes are moved away from the BERTVocab class. Conceptually it doesn't seem to make sense to expose them as part of the object, given that the object may not be using the default values. If you think this should be preserved, I can add it back.

szha:
@eric-haibin-lin @leezu if the attributes can be flexibly registered, do we still need BERTVocab?

leezu:
Only for backwards compatibility; I'm fine with breaking it, but we'll have to replace the Bert artifacts on S3

src/gluonnlp/vocab/vocab.py
```diff
@@ -203,6 +238,44 @@ def _index_counter_keys(self, counter, unknown_token, special_tokens, max_size,
...
+    def _sort_index_according_to_user_specification(self, token_to_idx):
+        # Sanity checks
+        if not set(token_to_idx.keys()).issubset(self.token_to_idx.keys()):
+            raise ValueError('User-specified token_to_idx mapping can only contain '
+                             'tokens that will be part of the vocabulary.')
+        if len(set(token_to_idx.values())) != len(token_to_idx):
+            raise ValueError('User-specified indices must not contain duplicates.')
+        if min(token_to_idx.values()) < 0 or max(token_to_idx.values()) >= len(self.token_to_idx):
+            raise ValueError('User-specified indices must not be < 0 or >= the number of tokens '
+                             'that will be in the vocabulary.')
```
eric-haibin-lin: Could you also include the value of `len(self.token_to_idx)` in the err msg? That helps the user to correct his idx value

src/gluonnlp/vocab/bert.py
```diff
@@ -97,113 +110,111 @@ class BERTVocab(Vocab):
...
-        return json.dumps(vocab_dict)
+            reserved_tokens = special_bert_tokens
+
+        if identifiers_to_tokens is None:
+            identifiers_to_tokens = dict()
+
+        if mask_token:
+            assert mask_token not in identifiers_to_tokens
+            identifiers_to_tokens.update(mask_token=mask_token)
+        if mask_token:
```
eric-haibin-lin: if sep_token

src/gluonnlp/vocab/bert.py
```diff
@@ -97,113 +110,111 @@ class BERTVocab(Vocab):
...
+            assert sep_token not in identifiers_to_tokens
+            identifiers_to_tokens.update(sep_token=sep_token)
+        if mask_token:
```
eric-haibin-lin: if cls_token

src/gluonnlp/vocab/bert.py
```diff
@@ -97,113 +110,111 @@ class BERTVocab(Vocab):
...
         ----------
         json_str : str
-            Serialized json string of a BERTVocab object.
+            Serialized json string of a Vocab object.
+
 
         Returns
         -------
-        BERTVocab
+        Vocab
```
eric-haibin-lin: Why not keep `BERTVocab ` as the return type in the doc?

src/gluonnlp/vocab/vocab.py
```diff
@@ -434,16 +507,26 @@ def from_json(cls, json_str):
...
+
+        vocab = cls(
+            counter=count_tokens(token_to_idx.keys()),
+            unknown_token=unknown_token,
+            padding_token=padding_token,
+            bos_token=bos_token,
+            eos_token=eos_token,
+            reserved_tokens=reserved_tokens,
+            token_to_idx=token_to_idx,
+            identifiers_to_tokens=vocab_dict.get('identifiers_to_tokens'))
```
eric-haibin-lin: This might not exist in previous serialized vocab files. Provide a default value?

src/gluonnlp/vocab/bert.py
```diff
@@ -97,113 +110,111 @@ class BERTVocab(Vocab):
...
+        special_tokens = [unknown_token, padding_token, bos_token, eos_token]
+        reserved_tokens = vocab_dict.get('reserved_tokens')
+
+        # workaround reserved and special tokens being serialized together
+        reserved_tokens = [
+            t for t in reserved_tokens if t not in special_tokens
+        ]
+
+        # Backwards compatibility with BERTVocab serialization in GluonNLP 0.7
+        identifiers_to_tokens = vocab_dict.get('identifiers_to_tokens')
```
eric-haibin-lin: This does not exist in previous vocab files

leezu: Do you mean the code did not exist? It is to enable loading the `BERTVocab` files that were created with GluonNLP 0.7; for example the ones distributed via S3. Or do you mean the 3 special tokens below did not exist in the serialized files?

src/gluonnlp/vocab/vocab.py
leezu: `dict.get` returns None if the value does not exist. Ie. the default value is None

leezu:
@eric-haibin-lin yes, backward compatibility is preserved. The `test_bertvocab` of https://github.com/dmlc/gluon-nlp/pull/509 checks it based on the models on S3. (We may want to improve it though to check the actual behaviour and not only the loading)

src/gluonnlp/vocab/bert.py
leezu: I meant GluonNLP 0.6

eric-haibin-lin: Oh I meant the `identifiers_to_tokens` field did not exist. can we provide a default value?

leezu:
Rebased after #749 was merged

mli:
Job PR-732/26 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-732/26/index.html

mli:
Job PR-732/27 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-732/27/index.html

mli:
Job PR-732/28 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-732/28/index.html

src/gluonnlp/embedding/token_embedding.py
```diff
@@ -334,23 +337,23 @@ def _load_embedding_serialized(self, pretrained_file_path):
...
                     vec_len = idx_to_vec.shape[1]
-                    idx_to_vec[C.UNK_IDX] = self._init_unknown_vec(shape=vec_len)
+                    idx_to_vec[UNK_IDX] = self._init_unknown_vec(shape=vec_len)
             else:
                 # If the TokenEmbedding shall not have an unknown token, we
                 # just delete the one in the npz.
-                assert C.UNK_IDX == 0
-                idx_to_token = deserialized_embedding.idx_to_token[C.UNK_IDX + 1:]
-                idx_to_vec = deserialized_embedding.idx_to_vec[C.UNK_IDX + 1:]
+                assert UNK_IDX == 0
```
szha: is this needed anymore?

src/gluonnlp/vocab/bert.py
```diff
@@ -70,25 +78,30 @@ class BERTVocab(Vocab):
...
-        A list of reserved tokens (excluding `unknown_token`) that will always be indexed, such as
-        special symbols representing padding, beginning of sentence, and end of sentence. It cannot
-        contain `unknown_token` or duplicate reserved tokens. Keys of `counter`, `unknown_token`,
-        and values of `reserved_tokens` must be of the same hashable type. Examples: str, int, and
-        tuple.
+    reserved_tokens : list of hashable objects or None, default None
+        A list specifying additional reserved tokens that will always be
+        indexed similar to `padding_token`, `bos_token` or `eos_token`.
+        `reserved_tokens` cannot contain `unknown_token` or duplicate reserved
+        tokens.
```
szha: what does it mean "indexed similar to other tokens"?

szha: the way that other tokens are indexed are not documented either.

src/gluonnlp/vocab/bert.py
```diff
@@ -70,25 +78,30 @@ class BERTVocab(Vocab):
...
+        Keys of `counter`, `unknown_token`, and values of `reserved_tokens`
+        must be of the same hashable type. Examples of hashable types are str,
+        int, and tuple.
+    identifiers_to_tokens : dict mapping elements of reserved_tokens to str or None, default None
```
szha: I'd suggest document how excessive tokens (i.e. tokens without a counter part in reserved_tokens) are handled and what happens to reserved tokens without an entry in this mapping.

Would it be possible to merge this with the reserved_tokens argument to avoid dealing with the mapping mismatch?

src/gluonnlp/embedding/token_embedding.py
leezu: Currently the TokenEmbedding uses a different methods to initialize it's internals depending on from what the user wants to construct the TokenEmbedding. It is thus a bit ugly to remove the UNK_IDX dependency and may introduce bugs easily.

The best way forward is to refactor the different place in TokenEmbedding that currently must modify and construct the internal attributes, making them static and extending the constructor to support initializing a TokenEmbedding given the list of tokens and the ndarray of embeddings.

I have prepared some code, but let's discuss it in a separate PR.

src/gluonnlp/vocab/bert.py
leezu: This should clarify "such as special symbols representing padding, beginning of sentence, and end of sentence" which was used before. Let's see if it can be made more clear.

src/gluonnlp/vocab/bert.py
leezu: Actually `identifiers_to_tokens` can specify identifiers for any token in the vocabulary, not just reserved tokens. I will correct the API doc. Only allowing to expose "valild" tokens that are part of the vocab as attributes acts as sanity check, by explicitly raising an error if the specified token is not valid, eg. due to a typo, instead of assuming that the new token must be added.

leezu: One downside of merging this with reserved_tokens is that it complicates the case where only some of the reserved tokens are exposed as attributes.

src/gluonnlp/vocab/bert.py
leezu: https://github.com/dmlc/gluon-nlp/pull/750

mli:
Job PR-732/29 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-732/29/index.html

src/gluonnlp/embedding/token_embedding.py
szha: Thanks for the explanation. I was referring to only the assert

src/gluonnlp/vocab/bert.py
szha: The original intention of reserved_tokens is for special tokens to be registered regardless of counts. If any regular token can be exposed as attribute, I'm not sure if we need the concept of reserved tokens anymore.

leezu:
@szha last commit introduces the registration of special tokens via keyword arguments as discussed. For deprecation of padding, bos, eos_tokens I'll open a separate PR.

You may want to take a look at the last commit @eric-haibin-lin

src/gluonnlp/vocab/bert.py
```diff
@@ -70,25 +78,23 @@ class BERTVocab(Vocab):
...
-    UNKNOWN_TOKEN : '[UNK]'
-        Static attribute of BERTVocab, unknown token for BERT.
-    PADDING_TOKEN : '[PAD]'
-        Static attribute of BERTVocab, padding token for BERT.
-    MASK_TOKEN : '[MASK]'
-        Static attribute of BERTVocab, mask token for BERT.
-    SEP_TOKEN : '[SEP]'
-        Static attribute of BERTVocab, a token used to separate sentence pairs for BERT.
-    CLS_TOKEN : '[CLS]'
-        Static attribute of BERTVocab, Classification symbol for BERT.
```
szha: Should these attributes still be documented? @eric-haibin-lin what do you think?

leezu: They are removed, so as long as they are not moved back, they shouldn't be documented. We can consider moving them back if you feel it's better

mli:
Job PR-732/32 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-732/32/index.html

PR #735, opened 2019-05-29
[MODEL] BERT conversion scripts, SciBERT, BioBERT, ClinicalBERT

leezu:
This refactors the tf Bert conversion scripts based on https://github.com/dmlc/gluon-nlp/pull/732. 

@eric-haibin-lin


leezu:
@eric-haibin-lin this PR only touches code in the script folder. All other changes are due to #732 and can be ignored while reviewing this.

scripts/bert/index.rst
```diff
@@ -294,7 +294,7 @@ Command line interface
 
 .. code-block:: shell
 
-    python bert_embedding/bert.py --sentences "GluonNLP is a toolkit that enables easy text preprocessing, datasets loading and neural models building to help you speed up your Natural Language Processing (NLP) research."
+    python bert/bert.py --sentences "GluonNLP is a toolkit that enables easy text preprocessing, datasets loading and neural models building to help you speed up your Natural Language Processing (NLP) research."
```
eric-haibin-lin: should it be `embedding.py`?

scripts/bert/compare_tf_gluon_model.py
```diff
@@ -129,8 +136,10 @@
 #                               Gluon MODEL                                   #
 ###############################################################################
 
+vocabulary = tf_vocab_to_gluon_vocab(tokenizer.vocab)
 bert, vocabulary = nlp.model.get_model(args.gluon_model,
-                                       dataset_name=args.gluon_dataset,
+                                       dataset_name=None,
+                                       vocab=vocabulary,
                                        pretrained=not args.gluon_parameter_file,
```
eric-haibin-lin: We have some converted TF model hosted on s3 that uses the old vocab. It seems that they won't pass the test due to difference in the vocab idx

scripts/bert/compare_tf_gluon_model.py
```diff
@@ -27,26 +27,33 @@
 import mxnet as mx
```
eric-haibin-lin: Could you mind creating a folder `convert` and moving 'compare_tf_model.py' and 'convert_tf_model.py' (and some necessary util functions) to that folder?

scripts/bert/compare_tf_gluon_model.py
leezu: The tests here are failing due to an old bug where duplicate special tokens (eg. `padding_token == eos_token`) would corrupt the `idx_to_token`. These corruptions are detected by some sanity checks in the updated flexible vocab implementation and tests were thus failing.
Backward compatibility for the Bert models should be preserved.

leezu: See https://github.com/dmlc/gluon-nlp/pull/749

mli:
Job PR-735/4 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-735/4/index.html

scripts/bert/conversion_tools/convert_tf_model.py
```diff
@@ -19,34 +19,49 @@
...
+                    help='Path to output folder. The folder must exist.')
 parser.add_argument('--debug', action='store_true', help='debugging mode')
 args = parser.parse_args()
 logging.getLogger().setLevel(logging.DEBUG if args.debug else logging.INFO)
 logging.info(args)
 
 # convert vocabulary
 vocab_path = os.path.join(args.tf_checkpoint_dir, 'vocab.txt')
-vocab, reserved_token_idx_map = convert_vocab(vocab_path)
+vocab = tf_vocab_to_gluon_vocab(load_tf_vocab(vocab_path))
```
eric-haibin-lin: This vocab is not compatible from the embedding of existing converted pre-trained models hosted on s3, right? Shall we use this only for new vocab like scibert/biobert?

leezu: What do you mean with compatible? Loading the serialized vocabs on S3 that were created with the previous code is supported, and they will be loaded correctly.
Running the `convert_tf_model.py` will not result in an identical Vocabulary though. In particular, unlike before, the sorting of the vocabulary index as given in the tensorflow vocab is preserved. But that shouldn't be necessary: The conversion output (vocab + parameters) are used together and we don't need to mix an old parameter with a new vocabulary.

Would there be any reason for re-generating the exact json as is currently on S3? If not, we should just remove the old code (as done in this PR) as it is 50 lines, compared to 5 lines now. Also the old code is preserved in the git history.

mli:
Job PR-735/5 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-735/5/index.html

scripts/bert/conversion_tools/convert_tf_model.py
eric-haibin-lin: I was thinking about the use case where we want to use the checkpoint and vocab previously distributed on s3 and validate them again with TF model. This is useful if we want to check if a particular change in mxnet causes difference in the model precision (e.g. layernorm or multi-head attention implementations are changed or optimized in mxnet). If so, we cannot use `tf_vocab_to_gluon_vocab ` with the existing checkpoint. 

On second thought, we don't necessarily need to rely on this script if we want to check if any operator change in mxnet causes differences. We can run BERT QA inference with a new mxnet commit to check if accuracy/F1 matches. 

For this PR, would you mind adding some documentations here that the vocabulary of this script is different from the one we distributed in gluonnlp 0.6?

leezu: But we can still use checkpoint and vocab that are previously distributed. The vocabulary used by tensorflow and the vocabulary used by gluon are independent. Or what do you mean?

I'm unsure what kind of documentation would be helpful here. The ordering of the vocabularies has never been documented / guaranteed, nor do people rely on it (the API is based on using the token_to_idx mapping, not based on hard coded indices). What use-case do you have in mind?

mli:
Job PR-735/7 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-735/7/index.html

scripts/bert/conversion_tools/convert_tf_model.py
eric-haibin-lin: If I use this script with `--gluon_dataset book_corpus_wiki_en_uncased --gluon_model bert_12_768_12 ` and leave `--gluon_parameter_file` as None, it would download the model from s3, generate a new vocab that has different indices from the one we host on s3, and the consistency check will fail, is that right? 

The embedding in the previous checkpoints assumes vocab[unk] = 0

leezu: Thanks @eric-haibin-lin for the elaboration.

If `--gluon_parameter_file` is not specified, the vocabulary from S3 should be used. But you're right that this was not correctly implemented in the comparison script. It is addressed in https://github.com/dmlc/gluon-nlp/pull/735/commits/59b225060503519ba9acef072a7a38380322bfdd

I added a `--gluon_vocab_file` argument, that is mandatory to specify if `--gluon_parameter_file` is given. It expects users to specify the vocab file generated by the `convert_tf_model.py` script they used to generate the param file. If no  `--gluon_parameter_file` is specified and the parameters are loaded from S3, the vocab will also be loaded from S3.

mli:
Job PR-735/8 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-735/8/index.html

mli:
Job PR-735/14 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-735/14/index.html

scripts/bert/conversion_tools/infer_pytorch_gluon_parameter_name_mapping.py
```diff
@@ -0,0 +1,96 @@
...
+                    help='Path to Tensorflow checkpoint folder.')
+parser.add_argument('--debug', action='store_true', help='debugging mode')
+parser.add_argument('--out', default='gluon_to_pytorch_naming.json',
+                    help='Output file to store gluon to pytorch name mapping.')
+args = parser.parse_args()
+logging.getLogger().setLevel(logging.DEBUG if args.debug else logging.INFO)
+logging.info(args)
+
+# Load Gluon Model
+# TODO root is only set until parameters are uploaded to S3
```
eric-haibin-lin: just a reminder to remove the root path

src/gluonnlp/data/utils.py
```diff
@@ -227,7 +227,16 @@ def _slice_pad_length(num_items, length, overlap=0):
...
                'wiki_cn': 'ddebd8f3867bca5a61023f73326fb125cf12b4f5',
                'wiki_multilingual_uncased': '2b2514cc539047b9179e9d98a4e68c36db05c97a',
-               'wiki_multilingual': '2b2514cc539047b9179e9d98a4e68c36db05c97a'}
+               'wiki_multilingual': '2b2514cc539047b9179e9d98a4e68c36db05c97a',
+               'scibert_scivocab_uncased': '2d2566bfc416790ab2646ab0ada36ba628628d60',
+               'scibert_scivocab_cased': '2c714475b521ab8542cb65e46259f6bfeed8041b',
+               'scibert_basevocab_uncased': '80ef760a6bdafec68c99b691c94ebbb918c90d02',
+               'scibert_basevocab_cased': 'a4ff6fe1f85ba95f3010742b9abc3a818976bb2c',
+               'biobert_v1.0_pmc': 'a4ff6fe1f85ba95f3010742b9abc3a818976bb2c',
+               'biobert_v1.0_pubmed': 'a4ff6fe1f85ba95f3010742b9abc3a818976bb2c',
```
eric-haibin-lin: For consistency, let's include cased/uncased in the suffiex?

scripts/bert/conversion_tools/convert_pytorch_model.py
```diff
@@ -0,0 +1,180 @@
...
+import sys
+
+import mxnet as mx
+import gluonnlp as nlp
+import torch
+from gluonnlp.model import BERTEncoder, BERTModel
+from gluonnlp.model.bert import bert_hparams
+
+sys.path.insert(0, os.path.abspath(os.path.join(__file__, os.pardir, os.pardir)))
+from utils import get_hash, load_tf_vocab, tf_vocab_to_gluon_vocab
```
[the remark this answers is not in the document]
leezu: It originates from the tf repo, but also used by the pytorch models. Let's rename it

src/gluonnlp/data/utils.py
leezu: The current names are the naming convention of the respective authors, which only published their models with a single vocabulary. There is for example no biobert_v1.0_pmc_cased and biobert_v1.0_pmc_uncased version. I think it's also ok to change the names as you suggest. It would require replacing the artifacts on S3.

mli:
Job PR-735/15 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-735/15/index.html

src/gluonnlp/data/utils.py
szha: done

mli:
Job PR-735/16 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-735/16/index.html

mli:
Job PR-735/17 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-735/17/index.html

PR #748, opened 2019-06-03
Fix conftest.py function_scope_seed

leezu:
`request.node.rep_call` is only present if the setup succeeded. Thus test for
`request.node.rep_setup.failed` before accessing `request.node.rep_call`.

This improves error messages during test development (ie. when running pytest
with a wrong test). Without this change, tests will fail with `AttributeError: 'Function' object has no attribute 'rep_call'`

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -177,7 +177,12 @@ def function_scope_seed(request):
 
     yield  # run the test
 
-    if request.node.rep_call.outcome == 'failed':
+    if request.node.rep_setup.failed:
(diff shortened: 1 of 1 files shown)


mli:
Job PR-748/1 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-748/1/index.html

leezu:
Codecoverage decreases as the added code will never execute as long as all tests are valid. Thus please ignore the decrease

Issue #789, opened 2019-06-21
Making transformer encoder fully hybridized for export


eric-haibin-lin:
Currently BERT can be [exported](https://github.com/dmlc/gluon-nlp/blob/master/scripts/bert/export/export.py) with static length support. BERTEncoder inherits Transformer encoder, which contains a few `.shape` API calls, making the transformer encoder not fully hybridizable and create issues during export. As the result, the exported model only supports static length. These calls are located here:
https://github.com/dmlc/gluon-nlp/blob/master/src/gluonnlp/model/transformer.py#L450-L463

We need to remove these calls to export a model that supports variable length. 

## handling arange
In particular, we have 
```
length = inputs.shape[1]
arange = mx.nd.arange(length, ctx=valid_length.context, dtype=valid_length.dtype)
```

To remove these .shape calls, we have 2 options:
### contrib.arange_like op with ndarray input
Instead of `contrib.arange(arr.shape[1], ...)`, we can introduce an arange_like op:
- input: `arr` with shape (x,) and abitrary data
- output: an output with shape `arr.shape`, and value of `[0, 1, 2, ... size(arr) - 1]`. 

With this op, we just need to slice the inputs on axis 1 and pass it to arange_like op: 
```
arr = inputs.slice(begin=(0,0,0), end=(0,None,0)
arange = F.contrib.arange(arr)
```

### control flow op

Alternatively, we can use control flow op (either foreach, or while loop) to loop N times, where N = inputs.shape[1]. Loop i fills in the value i in the output "arange" array.

However, this may have high overhead when N is large (512). 

## handling other `.shape` calls

- `mask = mx.nd.broadcast_axes(mx.nd.expand_dims(mask, axis=1), axis=1, size=length)` can be replaced with `broadcast_mul` op with `ones_like(arr)`
- `inputs * math.sqrt(inputs.shape[-1])` can be replace with `shape_nd` op

@TaoLv

pengxin99:
Just wonder, the op `arange`, `broadcast_axes`, `expand_dims` is all can be called by both NDArray and Symbol, so the all effort is to remove input.shape[] operator because symbol do not have `shape` operator?
thx @eric-haibin-lin @TaoLv 

BTW. we can use `export()` to get the hybrid model derictly from train_transformer.py .

nickguletskii:
The `contrib.index_array` operator could probably be used to replace these `shape` calls:

[...]

PR #845, opened 2019-07-24
[MODEL][API] Change BaseTransformerEncoder to HybridBlock

TaoLv:
## Description ##
This PR intends to change the BaseTransformerEncoder to HybridBlock so we can export it without padding to fixed sequence length.
https://github.com/dmlc/gluon-nlp/issues/789

TODOs:
- [x] branch if arange_like is not available
- [ ] change export.py under scripts/bert/export/

### Essentials ###
- [ ] Code is well-documented

### Changes ###

## Comments ##


src/gluonnlp/model/transformer.py
```diff
@@ -360,6 +361,13 @@ def __init__(self, attention_cell='multi_head', num_layers=2,
         self._scaled = scaled
         self._use_layer_norm_before_dropout = use_layer_norm_before_dropout
         self._scale_embed = scale_embed
+        self._support_arange_like = False
+        try:
+            self._support_arange_like = bool(ndarray.contrib.arange_like)
+        except AttributeError:
```
leezu: Other parts of the library will require mxnet 1.5 too. So you don't need provide a workaround for <1.5 IMO

sxjscience: I agree. We need to force the user to upgrade the version of MXNet.

eric-haibin-lin: this op is post 1.5 if i remember correctly

TaoLv: Yes, it's not in 1.5.0 release.

leezu: Can you use F.arange(..., infer_range=True) instead? See https://github.com/dmlc/gluon-nlp/pull/846/files#diff-deaa65ee7b0ea2b813a77275105de177R199-R200 for an example

TaoLv: Thank you @leezu. Does the below method look good to you? Since it will be sightly slower than the direct arange_like way, I will put it into the case when arange_like is not supported to replace the max_length way. @eric-haibin-lin 

```
import mxnet as mx
import numpy as np

a = mx.sym.Variable('a')
b = mx.sym.slice(a, begin=(0, 0, 0), end=(1, None, 1))
c = mx.sym.reshape(b, shape=(-1,))
d = mx.sym.zeros_like(c) + mx.sym.arange(start=0, stop=None, infer_range=True)

mod = d.simple_bind(mx.cpu(), a=(3,4,5), grad_req='null')
mod.arg_arrays[0][:] = np.random.normal(size=mod.arg_arrays[0].shape)
out = mod.forward(is_train=False)
print(out[0])
```

leezu: Yes, using `infer_range=True` to avoid the `max_length` requirement looks good.
Why do you mean it's slightly slower? Due to `zeros_like(c)`?

TaoLv: "arange_like" vs. "slice" + "zeros_like" + "elemwise_add". Here is a quick benchmark on my sever:

```
# shape = (32, 128, 1024)
infer_range true: 0.207459 ms
arange_like: 0.096955 ms
```

eric-haibin-lin:
We can limit this PR to hybridize transformer encoder only. I can take care of the hybridization of BERTModel and the export script in a separate PR.

TaoLv:
Thank you for working on this, @eric-haibin-lin . Sure, we can change the export script in another PR.

eric-haibin-lin:
Fixed by #877

PR #909, opened 2019-08-29
[FEATURE] Padding seq length to multiple of 8 in BERT model

ptrendx:
## Description ##
This PR adds an option to Pad batchify option to round the resulting size. It is especially important in FP16 training, like BERT training, as TensorCore algorithms may only be invoked when the sequence length is a multiple of 8.

@eric-haibin-lin FYI

### Essentials ###
- [x] Code is well-documented

### DIFF
--- a/scripts/bert/pretraining_utils.py
+++ b/scripts/bert/pretraining_utils.py
@@ -241,13 +241,13 @@ class BERTDataLoaderFn(DataLoaderFn):
         self._use_avg_len = use_avg_len
         self._num_ctxes = num_ctxes
         pad_val = vocab[vocab.padding_token]
-        self._batchify_fn = Tuple(Pad(pad_val=pad_val), # input_id
-                                  Pad(pad_val=pad_val), # masked_id
-                                  Pad(pad_val=0),       # masked_position
-                                  Pad(pad_val=0),       # masked_weight
-                                  Stack(),              # next_sentence_label
-                                  Pad(pad_val=0),       # segment_id
-                                  Stack())              # valid_length
+        self._batchify_fn = Tuple(Pad(pad_val=pad_val, round_to=8), # input_id
+                                  Pad(pad_val=pad_val),             # masked_id
+                                  Pad(pad_val=0),                   # masked_position
+                                  Pad(pad_val=0),                   # masked_weight
+                                  Stack(),                          # next_sentence_label
+                                  Pad(pad_val=0, round_to=8),       # segment_id
+                                  Stack())                          # valid_length
 
     def __call__(self, dataset, sampler):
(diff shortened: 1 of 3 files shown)


scripts/bert/pretraining_utils.py
```diff
@@ -241,13 +241,13 @@ def __init__(self, use_avg_len, num_ctxes, vocab):
...
         self._num_ctxes = num_ctxes
         pad_val = vocab[vocab.padding_token]
-        self._batchify_fn = Tuple(Pad(pad_val=pad_val), # input_id
-                                  Pad(pad_val=pad_val), # masked_id
-                                  Pad(pad_val=0),       # masked_position
-                                  Pad(pad_val=0),       # masked_weight
-                                  Stack(),              # next_sentence_label
-                                  Pad(pad_val=0),       # segment_id
-                                  Stack())              # valid_length
+        self._batchify_fn = Tuple(Pad(pad_val=pad_val, round_to=8), # input_id
```
TaoLv: Can it be controlled by an argument in command line?

mli:
Job PR-909/7 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-909/7/index.html

scripts/bert/pretraining_utils.py
eric-haibin-lin: We should enable it by default, i don't think an argument is necessary

eric-haibin-lin:
Merged. Thanks @ptrendx !

scripts/bert/pretraining_utils.py
szha: From description
> TensorCore algorithms may only be invoked when the sequence length is a multiple of 8.

TaoLv: Sorry I over looked that. So this script will never be executed on CPU or GPU without TensorCore?

leezu: @TaoLv good point. You could open a PR to add a cmd flag. If we can't auto-detect when TensorCore is present, it may still be best to enable the flag by default but give users the option to disable it.

eric-haibin-lin:
@ptrendx any idea how much performance difference would this leads to, typically?

ptrendx:
I recall from the BERT base pretraining experiments it was something like 20% for multi-GPU cases - basically the problem is that when you have multiple GPUs that need to sync with each other, having 1 GPU that hits the case where non-TensorCore algo is used leads to all other GPUs waiting for it.

eric-haibin-lin:
@ptrendx thanks for the clarification!

PR #685, opened 2019-04-30
[CI] address sequence sampler test problem

szha:
## Description ##
address sequence sampler test failure

### Essentials ###
- [ ] PR's title starts with a category (e.g. [BUGFIX], [MODEL], [TUTORIAL], [FEATURE], [DOC], etc)
- [ ] Changes are complete (i.e. I finished coding on this PR)
- [ ] All changes have test coverage
- [ ] Code is well-documented

### Changes ###
- [ ] Feature1, tests, (and when applicable, API doc)
- [ ] Feature2, tests, (and when applicable, API doc)

## Comments ##
- If this change is a backward incompatible change, why must this change be made.
- Interesting edge cases to note here

PR #1017, opened 2019-11-20
[CI] Remove skip_master feature

leezu:
## Description ##
Revert https://github.com/dmlc/gluon-nlp/pull/685
We are currently only running tests with the master version of MXNet.

Propose to remove this feature, as it is dangerous to disable tests with the `skip_master`, as we may forget to enable them again. Instead, we can add a conditional skip based on the current date and an estimate when the issue is fixed in MXNet master.

cc @szha


mli:
Job PR-1017/2 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-1017/2/index.html

Issue #726, opened 2019-05-24
race condition when horovod is used


eric-haibin-lin:
When launching multiple processes with horovod, downloading data may introduce race condition: 
```
  File "run_pretraining_hvd.py", line 255, in <module>
    start_step=args.start_step)
  File "/home/ec2-user/gluon-nlp/scripts/bert/pretraining_utils.py", line 44, in get_model_loss
    pretrained=pretrained, ctx=ctx)
  File "/home/ec2-user/.local/lib/python3.6/site-packages/gluonnlp-0.6.0-py3.6.egg/gluonnlp/model/__init__.py", line 151, in get_model
  File "/home/ec2-user/.local/lib/python3.6/site-packages/gluonnlp-0.6.0-py3.6.egg/gluonnlp/model/bert.py", line 612, in bert_24_1024_16
  File "/home/ec2-user/.local/lib/python3.6/site-packages/gluonnlp-0.6.0-py3.6.egg/gluonnlp/model/bert.py", line 680, in get_bert_model
  File "/home/ec2-user/.local/lib/python3.6/site-packages/gluonnlp-0.6.0-py3.6.egg/gluonnlp/model/utils.py", line 270, in _load_vocab
  File "/home/ec2-user/.local/lib/python3.6/site-packages/gluonnlp-0.6.0-py3.6.egg/gluonnlp/data/utils.py", line 310, in _load_pretrained_vocab
  File "/usr/lib64/python3.6/zipfile.py", line 1113, in __init__
    self.fp = io.open(file, filemode)
FileNotFoundError: [Errno 2] No such file or directory: '/home/ec2-user/.mxnet/models/book_corpus_wiki_en_cased-2d62af22.zip'
```

eric-haibin-lin:
```
[1,61]<stdout>:Downloading /home/ec2-user/.mxnet/models/1561012789.4379516book_corpus_wiki_en_uncased-a6607397.zip from https://apache-mxnet.s3-accelerate.dualstack.amazonaws.com/g$
uon/dataset/vocab/book_corpus_wiki_en_uncased-a6607397.zip...
[1,63]<stdout>:Downloading /home/ec2-user/.mxnet/models/1561012789.4379427book_corpus_wiki_en_uncased-a6607397.zip from https://apache-mxnet.s3-accelerate.dualstack.amazonaws.com/g$
uon/dataset/vocab/book_corpus_wiki_en_uncased-a6607397.zip...
[...]

PR #1078, opened 2019-12-22
[Parallel] Avoid race condition when downloading datasets

eric-haibin-lin:
## Description ##
Acquire a file lock before downloading the dataset. This helps when using multi-processing for multi-GPU training, or multi-machine training with a shared file system 

### Essentials ###
- [x] Code is well-documented

### Changes ###

## Comments ##

### DIFF
--- a/src/gluonnlp/data/utils.py
+++ b/src/gluonnlp/data/utils.py
@@ -21,13 +21,16 @@ import errno
 import os
 import tarfile
 import zipfile
-import time
+import random
+import sys
+import shutil
 
 import numpy as np
 from mxnet.gluon.data import SimpleDataset
 from mxnet.gluon.utils import _get_repo_url, check_sha1, download
 
 from .. import _constants as C
+from .. import utils
 
 __all__ = [
     'Counter', 'count_tokens', 'concat_sequence', 'slice_sequence', 'train_valid_split',
@@ -303,6 +306,11 @@ def _load_pretrained_vocab(name, root, cls=None):
     root = os.path.expanduser(root)
     file_path = os.path.join(root, file_name + '.vocab')
     sha1_hash = _vocab_sha1[name]
+
+    temp_num = str(random.Random().randint(1, sys.maxsize))
+    temp_root = os.path.join(root, temp_num)
+    temp_file_path = os.path.join(temp_root, file_name + '.vocab')
+    temp_zip_file_path = os.path.join(root, temp_num + file_name + '.zip')
     if os.path.exists(file_path):
         if check_sha1(file_path, sha1_hash):
             return _load_vocab_file(file_path, cls)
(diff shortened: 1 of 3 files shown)


eric-haibin-lin:
https://github.com/dmlc/gluon-nlp/issues/726

eric-haibin-lin:
Is there any recent change in numpy version on CI? The error does not seem to be relevant to the code changes.. @leezu

leezu:
Yes. Numpy 1.18 is released and one of our dependencies is incompatible.

leezu:
What kind of distributed filesystem are you using? It may be more efficient to cache the downloads on each machine locally. In my experience S3 is faster than for example EFS.
For multiprocessing, a workaround is to set the `MXNET_HOME` variable differently for each process?

I don't think MXNet upstream does locking / supports this? So introducing it only on the GluonNLP side may not be sufficient?

leezu:
> os.rename(): atomic but does not support across file system. 

I'm not sure how this is related to our use-case. Files in the same directory will always be in the same filesystem. Here you are interested in downloading a file and making it atomically available once the download is complete. This operation does not involve cross-filesystem renames.

> We do have lots of users using EFS/FSX.

It's a bad practice to keep the MXNET_HOME on EFS. EFS is slower than just using S3 or EBS. Also it is much more expensive. As a cheap and better performant alternative I suggest you look into https://github.com/kahing/goofys/

But in either way, this should be unrelated to atomic rename and locking.

eric-haibin-lin:
@leezu `os.lockf` is only available for Unix. I assume we don't want to stop supporting windows users. The advantage is that portalocker is portable across different operating systems. 

Yes, the support should be upstream'd to both gluon-nlp and mxnet

leezu:
Yes, if file locking is supported on Windows and we want to use file-locking, we should also support Windows.

So how about we first use idempotent operations in GluonNLP until support is upstreamed? Then we can reuse the upstream implementation.

sxjscience:
I think we can use the filelock. It's also used in sacrebleu: https://github.com/mjpost/sacreBLEU/blob/069b0c88fceb29f3e24c3c19ba25342a3e7f96cb/sacrebleu.py#L1154

eric-haibin-lin:
> So how about we first use idempotent operations in GluonNLP until support is upstreamed? Then we can reuse the upstream implementation.

What do you recommend as the way to implement idempotent operations and avoid race condition?

leezu:
File-renames within the same directory are idempotent and atomic. You mentioned above that rename is not supported across filesystems, but we don't need to cross-filesystem rename. The temporary file *with randomized name* should be in the same directory as the "target" file.

mli:
Job PR-1078/3 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-1078/3/index.html

mli:
Job PR-1078/4 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-1078/4/index.html

mli:
Job PR-1078/5 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-1078/5/index.html

PR #1087, opened 2020-01-03
[BUGFIX] fix the weight tied and weight sharing collision error

### LINKED ISSUE #1086: Conflict between weight tied and weight sharing
## Description
This bug is the same as https://github.com/apache/incubator-mxnet/issues/17184 It is an error introduced in gluonnlp models instead of mxnet side, so I move the issue here.

We experience some weight initialization error when we use weight sharing and weight tied simultaneously. We share weights between `model` and `model_eval`. The code is shown below:

```python
model = nlp.model.train.AWDRNN(args.model, len(vocab), args.emsize, args.nhid, args.nlayers,
                               args.tied, args.dropout, args.weight_dropout,
                               args.dropout_h, args.dropout_i, args.dropout_e)
model_eval = nlp.model.AWDRNN(args.model, len(vocab), args.emsize, args.nhid, args.nlayers,
                              args.tied, args.dropout, args.weight_dropout,
                              args.dropout_h, args.dropout_i, args.dropout_e,
                              params=model.collect_params())

model.initialize(mx.init.Xavier(), ctx=context)

model.hybridize(static_alloc=True)

print(model)

def check_initialized(net):
    params = net.collect_params()
    for param in params:
        try:
            params[param].list_ctx()
        except RuntimeError:
            return False
    return True

print(check_initialized(model))
print(check_initialized(model_eval))
```

### Log Message
If `args.tied` is set `True`, we get the following log message:
```python
True
False
```
If we turn off `args.tied`, the initialization works correctly.

### To Reproduce
(If you developed your own code, please provide a short script that reproduces the error. For existing examples, please provide link.)
The file can be found in (https://github.com/dmlc/gluon-nlp/blob/v0.8.x/scripts/language_model/word_language_model.py). To reproduce the above message, you may need to replace line 153 onward  with the above code snippet. Run the following command:

```
python -m pdb word_language_model.py --tied --dropout_e=0
```
You will encounter the above error.

## What have you tried to solve it?

The parameter that not initialized properly is the parameter `awdrnn0_hybridsequential0_embedding0_bias`. It is the weight used in the decoder of AWDRNN.  After some investigation, we found it is the tied weights introducing this error:

```python
    if self._tie_weights:
         output.add(nn.Dense(self._vocab_size, flatten=False,
                             params=self.embedding[0].params))
```

I print some debug information which may be helpful here:
```
(Pdb) model_eval.decoder[0]._params['awdrnn0_hybridsequential0_embedding0_bias'].list_ctx()
*** RuntimeError: Parameter 'awdrnn0_hybridsequential0_embedding0_bias' has not been initialized
```

## Environment

We recommend using our script for collecting the diagnositc information. Run the following command and paste the outputs below:
```
curl --retry 10 -s https://raw.githubusercontent.com/dmlc/gluon-nlp/master/tools/diagnose.py | python

# paste outputs here
```

liuzh91:
## Description ##
Resolve the conflict between weight sharing and weight tied. 
Fixes #1086

Running the code below:
```python
model = nlp.model.train.AWDRNN(args.model, len(vocab), args.emsize, args.nhid, args.nlayers,
                               args.tied, args.dropout, args.weight_dropout,
                               args.dropout_h, args.dropout_i, args.dropout_e)
model_eval = nlp.model.AWDRNN(args.model, len(vocab), args.emsize, args.nhid, args.nlayers,
                              args.tied, args.dropout, args.weight_dropout,
                              args.dropout_h, args.dropout_i, args.dropout_e,
                              params=model.collect_params())

model.initialize(mx.init.Xavier(), ctx=context)

model.hybridize(static_alloc=True)

print(model)

def check_initialized(net):
    params = net.collect_params()
    for param in params:
        try:
            params[param].list_ctx()
        except RuntimeError:
            return False
    return True

print(check_initialized(model))
print(check_initialized(model_eval))
```
It now produces the right answer:
```python
True
True
```

### Essentials ###
- [x] Code is well-documented

### Changes ###

## Comments ##

### DIFF
--- a/src/gluonnlp/model/train/language_model.py
+++ b/src/gluonnlp/model/train/language_model.py
@@ -18,7 +18,7 @@
 __all__ = ['AWDRNN', 'StandardRNN', 'BigRNN']
 
 from mxnet import init, nd, autograd
-from mxnet.gluon import nn, Block, HybridBlock, contrib, rnn
+from mxnet.gluon import nn, Block, HybridBlock, contrib, rnn, ParameterDict
 from mxnet import sym
 
 from ..utils import _get_rnn_layer, apply_weight_drop
@@ -71,6 +71,9 @@ class AWDRNN(HybridBlock):
         self._drop_e = drop_e
         self._weight_drop = weight_drop
         self._tie_weights = tie_weights
+        self._weight_share = False
+        if 'params' in kwargs and kwargs['params'] is not None:
+            self._weight_share = True
 
         with self.name_scope():
             self.embedding = self._get_embedding()
@@ -103,8 +106,15 @@ class AWDRNN(HybridBlock):
         output = nn.HybridSequential()
         with output.name_scope():
             if self._tie_weights:
-                output.add(nn.Dense(self._vocab_size, flatten=False,
-                                    params=self.embedding[0].params))
+                if self._weight_share:
+                    shared_params = self.embedding[0].params
+                    shared_params = ParameterDict(shared_params.prefix)
+                    shared_params.update(output._params._shared)
+                    output.add(nn.Dense(self._vocab_size, flatten=False,
+                                        params=shared_params))
+                else:
+                    output.add(nn.Dense(self._vocab_size, flatten=False,
+                                        params=self.embedding[0].params))
             else:
                 output.add(nn.Dense(self._vocab_size, flatten=False))
         return output
--- a/src/gluonnlp/model/train/language_model.py
+++ b/src/gluonnlp/model/train/language_model.py
@@ -71,6 +71,9 @@ class AWDRNN(HybridBlock):
         self._drop_e = drop_e
         self._weight_drop = weight_drop
         self._tie_weights = tie_weights
+        self._shared_params = None
+        if 'params' in kwargs:
+            self._shared_params = kwargs['params']
 
         with self.name_scope():
             self.embedding = self._get_embedding()
@@ -103,10 +106,10 @@ class AWDRNN(HybridBlock):
         output = nn.HybridSequential()
         with output.name_scope():
             if self._tie_weights:
-                if output._params._shared is not None:
+                if self._shared_params is not None:
                     shared_params = self.embedding[0].params
                     shared_params = ParameterDict(shared_params.prefix)
-                    shared_params.update(output._params._shared)
+                    shared_params.update(self._shared_params)
                     output.add(nn.Dense(self._vocab_size, flatten=False,
                                         params=shared_params))
                 else:
--- a/src/gluonnlp/model/train/language_model.py
+++ b/src/gluonnlp/model/train/language_model.py
@@ -71,9 +71,6 @@ class AWDRNN(HybridBlock):
         self._drop_e = drop_e
         self._weight_drop = weight_drop
         self._tie_weights = tie_weights
-        self._weight_share = False
-        if 'params' in kwargs and kwargs['params'] is not None:
-            self._weight_share = True
 
         with self.name_scope():
             self.embedding = self._get_embedding()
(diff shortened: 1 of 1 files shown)


mli:
Job PR-1087/1 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-1087/1/index.html

mli:
Job PR-1087/2 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-1087/2/index.html

leezu:
Please merge master due to #1096

mli:
Job PR-1087/4 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-1087/4/index.html

src/gluonnlp/model/train/language_model.py
```diff
@@ -71,6 +71,9 @@ def __init__(self, mode, vocab_size, embed_size=400, hidden_size=1150, num_layer
         self._drop_e = drop_e
         self._weight_drop = weight_drop
         self._tie_weights = tie_weights
+        self._shared_params = None
+        if 'params' in kwargs:
+            self._shared_params = kwargs['params']
```
leezu: I think your use-case is not well supported in the current Gluon API and the current solution may rely on some fragile behaviour in ParameterDict? 

The underlying problem is that `params=self.embedding[0].params` which is used in your PR and prior to your PR is not actually well supported currently due to the way that gluon handles parameter prefixes.

src/gluonnlp/model/train/language_model.py
```diff
@@ -103,8 +106,15 @@ def _get_decoder(self):
         output = nn.HybridSequential()
         with output.name_scope():
             if self._tie_weights:
-                output.add(nn.Dense(self._vocab_size, flatten=False,
-                                    params=self.embedding[0].params))
+                if self._shared_params is not None:
+                    shared_params = self.embedding[0].params
+                    shared_params = ParameterDict(shared_params.prefix)
+                    shared_params.update(self._shared_params)
```
leezu: What is the difference between `shared_params` and `self.embedding[0].params` after these 3 lines?

liuzh91: `shared_params` now shares the same prefix with `self.embedding[0].params`. `self._shared_params` has different `prefix` with `self.embedding[0].params`

liuzh91:
@leezu  the error is not due to the mishandle of prefix. It is triggered by the mismatch between `embedding[0].params` and `decoder[0].params`. To be concrete, `embedding[0]` has the following parameters
```
awdrnn0_hybridsequential0_embedding0_ (
  WeightDropParameter awdrnn0_hybridsequential0_embedding0_weight (shape=(33278, 400), dtype=float32, rate=0.1, mode=training)
)
```
while `decoder[0]` parameters are shown below:
```
awdrnn0_hybridsequential0_embedding0_ (
  WeightDropParameter awdrnn0_hybridsequential0_embedding0_weight (shape=(33278, 400), dtype=float32, rate=0.1, mode=training)
  Parameter awdrnn0_hybridsequential0_embedding0_bias (shape=(33278,), dtype=float32)
)
```
Weight tied only share the weight between `embedding` and `decoder`. It leaves parameter `bias` of the decoder uninitialized. Although `self._shared_params` contains all the shared parameters including `bias`from the training model, it cannot be directly passed as `params=self._shared_params` due to `prefix` mismatch.

leezu:
I see. Can you add a comment in the code?

liuzh91:
> I see. Can you add a comment in the code?

Done.

mli:
Job PR-1087/5 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-1087/5/index.html

PR #1115, opened 2020-01-15
FEATURE Bert fp16 norm: multi-tensor sum_sq

MoisesHer:
## Description ##
Make use of multi-tensor sum of squares during the computation of the norm in bert training.

### Essentials ###
- [x] Code is well-documented

### Changes ###
- [x] Instead of computing the dot product of each tensor sequentially, for later compute the sum of squares, in this PR we use the Mxnet operator multi_sum_sq, which computes the dot product of several tensors in parallel.
If multiple contexts/devices are available, the tensors are distributed among them, and later they are reduced into a single scalar.
- [x] Added test in test/test_utils.py (test_grad_global_norm)
- [x] Modified clip_grad_global_norm to use grad_global_norm

## Comments ##

### DIFF
--- a/scripts/bert/fp16_utils.py
+++ b/scripts/bert/fp16_utils.py
@@ -17,9 +17,7 @@
 
 """Trainer for mixed precision training."""
 import warnings
-import collections
 import mxnet as mx
-from mxnet import nd
 import gluonnlp as nlp
 
 class FP16Trainer:
--- a/scripts/bert/fp16_utils.py
+++ b/scripts/bert/fp16_utils.py
@@ -20,78 +20,7 @@ import warnings
 import collections
 import mxnet as mx
 from mxnet import nd
-from collections import defaultdict
-
-def grad_global_norm(parameters, max_norm):
-    """Calculate the 2-norm of gradients of parameters, and how much they should be scaled down
-    such that their 2-norm does not exceed `max_norm`.
-
-    If gradients exist for more than one context for a parameter, user needs to explicitly call
-    ``trainer.allreduce_grads`` so that the gradients are summed first before calculating
-    the 2-norm.
-
-    .. note::
-
-        This function is only for use when `update_on_kvstore` is set to False in trainer.
-
-    Example::
-
-        trainer = Trainer(net.collect_params(), update_on_kvstore=False, ...)
-        for x, y in mx.gluon.utils.split_and_load(X, [mx.gpu(0), mx.gpu(1)]):
-            with mx.autograd.record():
-                y = net(x)
-                loss = loss_fn(y, label)
-            loss.backward()
-        trainer.allreduce_grads()
-        norm, ratio = grad_global_norm(net.collect_params().values(), max_norm)
-        trainer.update(batch_size * ratio)
-        ...
-
-    Parameters
-    ----------
-    parameters : list of Parameters
-
-    Returns
-    -------
-    NDArray
-      Total norm. Shape is (1,)
-    NDArray
-      Ratio for rescaling gradients based on max_norm s.t. grad = grad / ratio.
-      If total norm is NaN, ratio will be NaN, too. Shape is (1,)
-    NDArray
-      Whether the total norm is finite. Shape is (1,)
-    """
-    # distribute gradients among contexts
-    idx = 0
-    arrays = defaultdict(list)
-    sum_norms = []
-    for p in parameters:
-        if p.grad_req != 'null':
-            p_grads = p.list_grad()
-            arrays[idx % len(p_grads)].append(p_grads[idx % len(p_grads)])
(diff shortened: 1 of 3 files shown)


mli:
Job PR-1115/1 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-1115/1/index.html

scripts/bert/fp16_utils.py
```diff
@@ -20,6 +20,7 @@
 import collections
 import mxnet as mx
 from mxnet import nd
+from collections import defaultdict
 
 def grad_global_norm(parameters, max_norm):
     """Calculate the 2-norm of gradients of parameters, and how much they should be scaled down
```
eric-haibin-lin: Actually, would you mind helping move this function to src/gluonnlp/utils/parameter.py?

API Spec:
```
def grad_global_norm(parameters, max_norm=None):
    """Calculate the 2-norm of gradients of parameters, and how much they should be scaled down
    such that their 2-norm does not exceed `max_norm`, if `max_norm` if provided.

    If gradients exist for more than one context for a parameter, user needs to explicitly call
    ``trainer.allreduce_grads`` so that the gradients are summed first before calculating
    the 2-norm.

    .. note::
        This function is only for use when `update_on_kvstore` is set to False in trainer.

    Example::

        trainer = Trainer(net.collect_params(), update_on_kvstore=False, ...)
        for x, y in mx.gluon.utils.split_and_load(X, [mx.gpu(0), mx.gpu(1)]):
            with mx.autograd.record():
                y = net(x)
                loss = loss_fn(y, label)
            loss.backward()
        trainer.allreduce_grads()
        norm = grad_global_norm(net.collect_params().values())
        ...

    Parameters
    ----------
    parameters : list of Parameters
    max_norm: NDArray, optional
        The maximum L2 norm threshold. If provided, `ratio` and `is_finite` will be returned.

    Returns
    -------
    NDArray
      Total norm. Shape is (1,)
    NDArray
      Ratio for rescaling gradients based on max_norm s.t. grad = grad / ratio.
      If total norm is NaN, ratio will be NaN, too.
      Returned if `max_norm` is provided. Shape is (1,)
    NDArray
      Whether the total norm is finite, returned if `max_norm` is provided. Shape is (1,)
    """

```

Then we can add an unit test for `gluonnlp.utils.grad_global_norm`. We can update the implementation of `gluonnlp.utils.clip_grad_global_norm` to call `grad_global_norm` first, too.

leezu:
@MoisesHer please fix lint

src/gluonnlp/utils/parameter.py
```diff
@@ -16,16 +16,89 @@
...
+    assert len(arrays) > 0, 'No parameter found available for gradient norm.'
+
+    ctx, dtype = arrays[0][0].context, 'float32'
+    for idx, arr in enumerate(arrays.values()):
+        sum_norm = mx.nd.multi_sum_sq(*arr, num_arrays=len(arr))
+        sum_norm = nd.add_n(*sum_norm)
+        sum_norms.append(sum_norm.as_in_context(ctx))
+
+    # reduce
+    total_norm = nd.add_n(*sum_norms).sqrt()
```
eric-haibin-lin: could you return here if max_norm is None?

eric-haibin-lin: And add a test, too?

MoisesHer: > could you return here if max_norm is None?
Yeap, 
What values should I return for (ratio, is_finite) ?

mli:
Job PR-1115/2 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-1115/2/index.html

src/gluonnlp/utils/parameter.py
eric-haibin-lin: just `return total_norm`

mli:
Job PR-1115/3 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-1115/3/index.html

src/gluonnlp/utils/parameter.py
```diff
@@ -16,16 +16,90 @@
...
+    if not max_norm:
```
eric-haibin-lin: can we change this to `if max_norm is None`? The difference is that, `if not max_norm` will evaluate the statement, implicitly invoking asscalar() on `max_norm` which blocks the python main thread.

mli:
Job PR-1115/4 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-1115/4/index.html

PR #1123, opened 2020-01-19
[Feature] Add Ner Suffix feature

XINGXIAOYU:
## Description ##
Add a parameter "tagging_first_token", so you can choose to use the first piece or the last piece of each word. The first piece catches the prefix feature of a word, and the last piece catches the suffix feature of a word.

### Essentials ###
- [x] Code is well-documented

## Comments ##

### DIFF
--- a/scripts/sequence_labeling/data.py
+++ b/scripts/sequence_labeling/data.py
@@ -131,7 +131,7 @@ def remove_docstart_sentence(sentences):
     return ret
 
 
-def bert_tokenize_sentence(sentence, bert_tokenizer):
+def bert_tokenize_sentence(sentence, bert_tokenizer, tagging_first_token):
     """Apply BERT tokenizer on a tagged sentence to break words into sub-words.
     This function assumes input tags are following IOBES, and outputs IOBES tags.
 
@@ -141,6 +141,9 @@ def bert_tokenize_sentence(sentence, bert_tokenizer):
         List of tagged words
     bert_tokenizer: nlp.data.BertTokenizer
         BERT tokenizer
+    tagging_first_token: bool, optional (default: True)
+        By default, only the first token of a word is going to be tagged.
+        If ``tagging_first_token`` is set to False, then the last token of a word is going to be tagged.
 
     Returns
     -------
@@ -151,14 +154,20 @@ def bert_tokenize_sentence(sentence, bert_tokenizer):
         # break a word into sub-word tokens
         sub_token_texts = bert_tokenizer(token.text)
         # only the first token of a word is going to be tagged
-        ret.append(TaggedToken(text=sub_token_texts[0], tag=token.tag))
-        ret += [TaggedToken(text=sub_token_text, tag=NULL_TAG)
-                for sub_token_text in sub_token_texts[1:]]
+        if tagging_first_token:
+            ret.append(TaggedToken(text=sub_token_texts[0], tag=token.tag))
+            ret += [TaggedToken(text=sub_token_text, tag=NULL_TAG)
+                    for sub_token_text in sub_token_texts[1:]]
+        # only the last token of a word is going to be tagged
+        else:
+            ret += [TaggedToken(text=sub_token_text, tag=NULL_TAG)
+                    for sub_token_text in sub_token_texts[:-1]]
+            ret.append(TaggedToken(text=sub_token_texts[-1], tag=token.tag))
 
     return ret
 
 
-def load_segment(file_path, bert_tokenizer):
+def load_segment(file_path, bert_tokenizer, tagging_first_token):
     """Load CoNLL format NER datafile with BIO-scheme tags.
 
     Tagging scheme is converted into BIOES, and words are tokenized into wordpieces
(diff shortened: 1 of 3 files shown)


mli:
Job PR-1123/1 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-1123/1/index.html

scripts/sequence_labeling/finetune_bert.py
```diff
@@ -77,6 +77,8 @@ def parse_args():
                             help='Learning rate for optimization')
     arg_parser.add_argument('--warmup-ratio', type=float, default=0.1,
                             help='Warmup ratio for learning rate scheduling')
+    arg_parser.add_argument('--tagging-first-token', type=str2bool, default=True,
```
leezu: How about `parser.add_argument('--tag-last-token', action='store_true')`. It seems simpler to call `finetune_bert.py --tag-last-token` than `finetune_bert.py --tagging-first-token=False`.

In either case please update the test case in `scripts/tests/` to run invoke the `finetune_bert.py` with both options. You can parametrize the test following for example Haibin's recent PR: https://github.com/dmlc/gluon-nlp/pull/1121/files#diff-fa82d34d543ff657c2fe09553bd0fa34R234

XINGXIAOYU: Sure, I will update it.

sxjscience:
Have you found any performance differences?

XINGXIAOYU:
@sxjscience I've tried the default parameters set in the scripts on conll2003 dataset. The performance using suffix feature will be a little lower than using the prefix feature.

sxjscience:
I think we can try the following:
- use the state corresponds to the first subword token
- use the state corresponds to the last subword token
- use the average pooling on top of the states of all the subwords
- use the max pooling on top of the states of all the subwords

sxjscience:
One problem is that since we are using self-attention, we are able to tailor the attention weights to cover the first, last, average cases. Thus, I don't think selecting the first/last token will impact the performance much.

XINGXIAOYU:
@sxjscience In classification task, I think it does not matter. But in sequence labeling task, one word has one label. If we break the word 'w' into several subwords [sw1,sw2,...], then only sw1 will have the label, and the labels of the others will set to NULL. I think it does not make sense.

scripts/sequence_labeling/predict_ner.py
```diff
@@ -81,7 +85,7 @@ def main(config):
                                             train_config.dropout_prob)
 
     dataset = BERTTaggingDataset(text_vocab, None, None, config.test_path,
-                                 config.seq_len, train_config.cased, tag_vocab=tag_vocab)
+                                 config.seq_len, train_config.cased, tag_vocab=tag_vocab,tagging_first_token=config.tagging_first_token)
```
liuzh91: Pls add white space after the comma.

sxjscience:
Due to the fact that we are using attention, the state bound to sw1 will be related to the other sub-words. The same thing happens for sw_n. A reasonable approach is to mask the loss corresponding to the other sub-word tokens and only use the state of the first subword as the contextualized word embedding.

scripts/sequence_labeling/data.py
```diff
@@ -337,19 +358,32 @@ def convert_arrays_to_text(text_vocab, tag_vocab,
...
-
+                if true_tag == NULL_TAG:
+                    tmptext += token_text
+                else:
+                    if len(tmptext) > 0:
+                        text = tmptext + token_text
+                        entries.append(PredictedToken(text=text,
+                                                      true_tag=true_tag, pred_tag=pred_tag))
+                        tmptext = ''
+                    else:
```
liuzh91: Can both cases be merged here? For example, if `len(tmptext) == 0`, you can still have `text = tmptext + token_text` which is equivalent to `token_text`.

scripts/sequence_labeling/data.py
```diff
@@ -337,19 +358,32 @@ def convert_arrays_to_text(text_vocab, tag_vocab,
...
+                                                 pred_tag=last_entry.pred_tag)
+                else:
+                    entries.append(PredictedToken(text=token_text,
+                                                  true_tag=true_tag, pred_tag=pred_tag))
             else:
-                entries.append(PredictedToken(text=token_text,
-                                              true_tag=true_tag, pred_tag=pred_tag))
-
+                if true_tag == NULL_TAG:
+                    tmptext += token_text
```
liuzh91: better name it as `tmp_text`. Or what about `partial_text`?

XINGXIAOYU:
@sxjscience Agree with you, and I'll try this method.

liuzh91:
> A reasonable approach is to mask the loss corresponding to the other sub-word tokens and only use the state of the first subword as the contextualized word embedding.

I am confused about this part. Why masking loss of other sub-word tokens is reasonable? For example on NER tasks, suffix is much more important than prefix of the words like `firefighter`.

sxjscience:
> > A reasonable approach is to mask the loss corresponding to the other sub-word tokens and only use the state of the first subword as the contextualized word embedding.
> 
> I am confused about this part. Why masking loss of other sub-word tokens is reasonable? For example on NER tasks, suffix is much more important than prefix of the words like `firefighter`.

Since we are using attention, the higher-level state associated with `fire` has already captured the information about `fighter`. Suffix should be more appropriate if we are using a LSTM. However, we are using attention and we are actually attending to the whole input.

StrayBird-ATSH:
@sxjscience Do you think we should continue with this pull request?
If we should, maybe I can take over the remaining work that needs to be done on this pull request.

PR #1152, opened 2020-02-11
Update master build URLs

leezu:
## Description ##
The URL scheme recently changed. CC @szha


mli:
Job PR-1152/1 is complete. 
Docs are uploaded to http://gluon-nlp-staging.s3-accelerate.dualstack.amazonaws.com/PR-1152/1/index.html

szha:
note that this is temporary because the versioning for the binary distribution is done incorrectly for convenience. the url will need to be corrected again later.

PR #1235, opened 2020-05-27
trigger build

szha:
## Description ##
(Brief description on what this PR is about)

### Essentials ###
- [ ] Code is well-documented

### Changes ###

## Comments ##

cc @dmlc/gluon-nlp-team

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -129,6 +129,7 @@ For advanced examples, check out our
 For experienced users, check out our
 `API Notes <http://gluon-nlp.mxnet.io/master/api/index.html>`__.
 
+
 Quick Start Guide
 =================
 


StrayBird-ATSH:
Currently, in `src/gluonnlp/data/conll.py` Line 143, we are using URL `http://www.clips.uantwerpen.be/conll2000/chunking/`.
However, as I try to visit this website, it is redirecting to `https://www.clips.uantwerpen.beconll2000/chunking/`, and the redirected server cannot be found.

I think this is the root cause of the failure because this action downloads a dataset, without which the following test cannot be run.

I think we need to have our own copy of the dataset on aws storage to avoid any other future failures that are due to external server faults.

PR #1249, opened 2020-06-13
[Numpy] Add "match_tokens_with_char_spans" + Enable downloading from S3 + Add Ubuntu test

sxjscience:
- [x] Add match_tokens_with_char_spans to utility
We convert the character spans to token starts+ends based on binary search
- [x] Enable downloading from S3. Now we are able to call
```python
from gluonnlp.utils import download
download('s3://commoncrawl/crawl-data/CC-MAIN-2020-24/segments/1590347385193.5/wet/CC-MAIN-20200524210325-20200525000325-00003.warc.wet.gz', overwrite=True)
```
- [x] Add Ubuntu test

In order for this feature to work, the user needs to configure S3 correctly.

Also, speed test shows that downloading from S3 can be around 4 times faster in a c4.8x machine in EC2:
```python
In [7]: %timeit download('s3://commoncrawl/crawl-data/CC-MAIN-2020-24/segments/1590347385193.5/
   ...: wet/CC-MAIN-20200524210325-20200525000325-00003.warc.wet.gz', overwrite=True)
Downloading CC-MAIN-20200524210325-20200525000325-00003.warc.wet.gz from s3://commoncrawl/crawl-data/CC-MAIN-2020-24/segments/1590347385193.5/wet/CC-MAIN-20200524210325-20200525000325-00003.warc.wet.gz...
100%|██████████████████████████████████████████████████████| 157M/157M [00:01<00:00, 87.3MiB/s]
2.5 s ± 401 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [8]: %timeit download('https://commoncrawl.s3.amazonaws.com/crawl-data/CC-MAIN-2020-24/segme
   ...: nts/1590347385193.5/wet/CC-MAIN-20200524210325-20200525000325-00003.warc.wet.gz', overw
   ...: rite=True)
Downloading CC-MAIN-20200524210325-20200525000325-00003.warc.wet.gz from https://commoncrawl.s3.amazonaws.com/crawl-data/CC-MAIN-2020-24/segments/1590347385193.5/wet/CC-MAIN-20200524210325-20200525000325-00003.warc.wet.gz...
100%|██████████████████████████████████████████████████████| 157M/157M [00:08<00:00, 19.0MiB/s]
8.13 s ± 389 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

This will help us to download large datasets like wikipedia + commoncrawl.


sxjscience:
@ZheyuYe We may try to revise our wikipedia downloading script as:
1) Try to use S3
2) Fallback to https if S3 raises an exception.

sxjscience:
Find that wikipedia is not available in S3. However, CommonCrawl is in S3 so this functionality is helpful for us to downloading commoncrawl.

ZheyuYe:
Perhaps it is also possible to fix some  invalid links in datasets in this PR like **General NLP Benchmarks** and **scripts** in like https://github.com/dmlc/gluon-nlp/blob/numpy/scripts/datasets/README.md

sxjscience:
I'm not sure why it gives s `61.40%` diff hit. I'll merge this in first. The test coverage is actually increased after this PR.

codecov.yml
```diff
@@ -0,0 +1,45 @@
...
+        paths:
+          - scripts/datasets
+      preprocess_cli:
+        target: auto
+        paths:
+          - scripts/preprocess
+    patch:
+      default:
+        target: 50%
+        base: auto
```
szha, line 29: let's not change this to a lower number.

PR #1254, opened 2020-07-02
Update codecov action to handle different OS and Python versions

leezu:
https://github.com/codecov/codecov-action/issues/80#issuecomment-645189406


szha:
how do we move this forward?

PR #1264, opened 2020-07-15
Add fp16 support for Bert QA inference

MoisesHer:
## Description ##
Adding FP16 in QA-BERT inference

### Essentials ###
- [x] Code is well-documented

### Changes ###
- [x] Added option to use fp16 data type in run_squad script for evaluation (--eval_dtype). Will cast the network.
- [x] Propagated the datatype to masked_softmax, where was using fp32 always and causing nans if fp16 was selected

## Comments ##

cc @dmlc/gluon-nlp-team, @sxjscience

### DIFF
--- a/scripts/question_answering/run_squad.py
+++ b/scripts/question_answering/run_squad.py
@@ -119,6 +119,8 @@ def parse_args():
                         help='Whether to evaluate all intermediate checkpoints instead of only last one')
     parser.add_argument('--max_saved_ckpt', type=int, default=10,
                         help='The maximum number of saved checkpoints')
+    parser.add_argument('--eval_dtype', type=str, default='float32',
+                        help='Data type used for evaluation. Either float32 or float16')
     args = parser.parse_args()
     return args
 
@@ -285,7 +287,8 @@ def get_network(model_name,
                 ctx_l,
                 dropout=0.1,
                 checkpoint_path=None,
-                backbone_path=None):
+                backbone_path=None,
+                dtype='float32'):
     """
     Get the network that fine-tune the Question Answering Task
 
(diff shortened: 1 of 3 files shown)


scripts/question_answering/run_squad.py
```diff
@@ -311,13 +314,14 @@ def get_network(model_name,
     # Create the network
     Model, cfg, tokenizer, download_params_path, _ = \
         get_backbone(model_name, load_backbone=not backbone_path)
-    backbone = Model.from_cfg(cfg, use_pooler=False)
+    backbone = Model.from_cfg(cfg, use_pooler=False, dtype=dtype)
```
sxjscience, line 317: Would you try to use:
```python
cfg.defrost()
cfg.MODEL.dtype = dtype
cfg.freeze()

backbone = Model.from_cfg(cfg, use_pooler=False)
```
So there is no need to revise the cfg parsing logic.

scripts/question_answering/run_squad.py
```diff
@@ -777,7 +784,9 @@ def predict_extended(original_feature,
 def evaluate(args, last=True):
     ctx_l = parse_ctx(args.gpus)
     cfg, tokenizer, qa_net = get_network(
-        args.model_name, ctx_l, args.classifier_dropout)
+        args.model_name, ctx_l, args.classifier_dropout, dtype=args.eval_dtype)
+    if args.eval_dtype == 'float16':
+        qa_net.cast('float16')
```
sxjscience, line 789: Is it possible to merge this into `get_network`? We may add tests in a later PR.

scripts/question_answering/run_squad.py
```diff
@@ -311,13 +314,17 @@ def get_network(model_name,
     # Create the network
     Model, cfg, tokenizer, download_params_path, _ = \
         get_backbone(model_name, load_backbone=not backbone_path)
+    cfg.defrost()
+    cfg.MODEL.dtype = dtype
+    cfg.freeze()
```
szha, line 319: changing it here doesn't seem like a good general approach. would it be better to add the datatype support to `get_backbone`?

sxjscience: After a second thought. May be it's better to add dtype to `get_backbone`.

scripts/question_answering/run_squad.py
MoisesHer: in the case of training, I think we do not want to cast if eval_type is fp16, that is the reason to do it here.
In the coming days I think we can AMP for training as well

scripts/question_answering/run_squad.py
MoisesHer: ok, np, I made the change back

sxjscience:
Will need to add test later.

PR #1270, opened 2020-07-18
Fix BERT fp16 bugs, add test

MoisesHer:
## Description ##
Fixing fp16 bug in a previous PR: https://github.com/dmlc/gluon-nlp/pull/1264
A bug was introduced in the last commit (merging with numpy upstream) of that PR when solving a conflict. Apologies
Bug only affects fp16 case, returning NaNs
Model was also not hybridized after being cast.
[x] Both issues are solved in this PR.
[x] A test was added to compare FP32 vs. FP16 results in BERT inference

cc @dmlc/gluon-nlp-team @sxjscience


szha:
Could you add a test for the expected data type?

sxjscience:
@MoisesHer Would you help add a test in https://github.com/dmlc/gluon-nlp/blob/numpy/tests/test_models_bert.py? You can do a similar forward test as in https://github.com/dmlc/gluon-nlp/blob/e78a24eeb124ff9efb99b1aee7ab1ee9181c9e0f/tests/test_models_roberta.py#L16-L48 and compare the final result.

sxjscience:
Need to wait for the GPU CI.

sxjscience:
@MoisesHer The GPU CI should be functional now. Would you try to create a new PR to add FP16 functionality?

sxjscience:
You can refer to https://github.com/dmlc/gluon-nlp/blob/a842bf3c0e33bbcfc376f57ea6e5bccda4144a55/tests/test_models_bert.py#L15 . Here, we just add a `ctx` to the arguments so it becomes a fixture.

sxjscience (approved):
LGTM

sxjscience:
CC @ZheyuYe @szha @hymzoque I'll merge this in first.

PR #1276, opened 2020-07-24
Horovod support for pretraining and fune-tuning squad 

ZheyuYe:
## Description ##
Thanks @eric-haibin-lin for providing Horovod support in MXNet 2.0,  see https://github.com/eric-haibin-lin/horovod/tree/mx2. Based on this, a sight  modification were made to support the use of Horovod in the GluonNLP number version.

## Changes ##
- [x] Refactor the `prepare_wikipedia.py` scripts with multiprocessing and uplaod the 2020-06-20 version to S3
- [x] Ensure the structural consistency of the pre-processing corpus which can be directly loaded into the pre-training script as `--data 'pretraining_data/pretraining_data/*.train,pretraining_data/prepared_wikipedia/*.txt'` 
- [x] Refactor the pretraining and squad fune-tuning scripts to support horovod
- [x] Move layer-wise learning rate decay and froze parameters approch into backbone model (electra)
- [x] Speed up electra with the help of `mx.npx.index_update` implemented in https://github.com/apache/incubator-mxnet/pull/18545
- [x] upload gluon_electra_small_owt to S3

## Comments ##
@sxjscience  @hymzoque


scripts/question_answering/README.md
```diff
@@ -70,6 +70,13 @@ python run_squad.py \
     --overwrite_cache \
 ```
 
+We support multi-GPU training via horovod:
```
leezu, line 73: Would you like to include performance numbers here, pointing out how much speedup users can expect by using horovod?

ZheyuYe: Sure, the  performance numbers may quite different depend on which model we are training.  I only included the numbers of roberta large here.

src/gluonnlp/models/electra.py
```diff
@@ -337,6 +347,58 @@ def get_initial_embedding(self, F, inputs, token_types=None):
...
+            layer-wise decay power
+        not_included: list of str
+            A list or parameter names that not included in the layer-wise decay
+        """
+
+        # consider the task specific finetuning layer as the last layer, following with pooler
+        # In addition, the embedding parameters have the smaller learning rate based on this setting.
+        max_depth = self.num_layers
+        for _, value in self.collect_params('.*embed*').items():
+            value.lr_mult = layerwise_decay**(max_depth + 1)
```
sxjscience, line 368: For me, I find that I need to use `max_depth = self.num_layers + 2` for text classification: https://github.com/sxjscience/gluon-nlp/blob/adcd22301f73ea05aa366bbb41c3016f3aac5186/src/gluonnlp/auto/estimators/basic_v1.py#L93-L108

src/gluonnlp/models/electra.py
```diff
@@ -337,6 +347,58 @@ def get_initial_embedding(self, F, inputs, token_types=None):
         embedding = self.embed_dropout(embedding)
         return embedding
 
+    def apply_layerwise_decay(self, layerwise_decay, not_included=[]):
```
sxjscience, line 350: Usually, we need to avoid setting default value to list. Instead, we may use a None.

scripts/question_answering/README.md
leezu: Thank you for adding it. It will still provide a rough estimate to users

scripts/datasets/pretrain_corpus/prepare_wikipedia.py
```diff
@@ -145,32 +170,75 @@ def download_wikicorpus(lang, date, output):
...
+
+    # Merge extracted content into txt files
+    prepared_path = os.path.join(output, 'prepared_wikipedia')
+    if not os.path.exists(prepared_path):
+        os.makedirs(prepared_path)
+    filenames = get_formatting_list(wiki_path, recursive=True)
+    num_files = len(filenames)
+    num_out_files = min(num_out_files, num_files)
+    file_volume = math.ceil(num_files / num_out_files)
+    splited_files = [filenames[i: i + file_volume] for i in range(0, num_files, file_volume)]
```
sxjscience, line 195: Instead of doing this, you may use `np.array_split`: See document https://numpy.org/doc/stable/reference/generated/numpy.array_split.html

scripts/pretraining/README.md
```diff
@@ -3,25 +3,27 @@
 Following the instruction of [Prepare OpenWebTextCorpus](../datasets/pretrain_corpus#openwebtext), download and prepare the dataset, obtaining a total of 20610 text files in the folder `prepared_owt`.
 
 ```bash
-python preprocesse_owt.py --input prepared_owt --output preprocessed_owt --shuffle
+python data_preprocessing.py --input prepared_owt --output preprocessed_owt --max_seq_length 128
```
sxjscience, line 6: Still prefer to add shuffle here.

scripts/pretraining/data_preprocessing.py
```diff
@@ -8,8 +8,12 @@
 import multiprocessing
 
 from pretraining_utils import get_all_features
+from gluonnlp.base import get_repo_model_zoo_url
+from gluonnlp.utils.misc import download
 from gluonnlp.data.tokenizers import HuggingFaceWordPieceTokenizer
 
+VOCAB_PATH = 'google_electra_small/vocab-e6d2b21d.json'
```
sxjscience, line 15: I suggest adding name of the backbone model so it will use the same tokenizer as the backbone.

ZheyuYe: Could we have a  function called `get_tokenizer` that plays similar role as `get_backbone` but only return tokenizer and vocab according to the given backbone name

sxjscience: Yes, we can.

sxjscience: In fact, it may just be a thing wrapper around get_backbone because each backbone has a tokenizer.

sxjscience:
@szhengac @ZiyueHuang This is merged and you may try pretraining with Electra + Horovod

scripts/pretraining/pretraining_utils.py
```diff
@@ -496,13 +494,17 @@ def dynamic_masking(self, F, input_ids, valid_lengths):
         valid_candidates = valid_candidates.astype(np.float32)
         num_masked_position = F.np.maximum(
             1, F.np.minimum(N, round(valid_lengths * self._mask_prob)))
-        # The categorical distribution takes normalized probabilities as input
-        # softmax is used here instead of log_softmax
```
ZheyuYe, line 500: `log(sample_probs)` instead of `sample_prob`s should be used for `npx.topk `

PR #1280, opened 2020-07-29
[Numpy] Fix SQuAD + Fix GLUE downloading

sxjscience:
## Description ##
Fix SQuAD due to the recent upgrade of MXNet: https://github.com/apache/incubator-mxnet/pull/18368 @MoisesHer Basically, the default data type of integer arrays will be int64.

### Changes ###
- [x] SQuAD Fix

cc @dmlc/gluon-nlp-team


szha:
thanks for bringing this up. int64 as default is not yet enabled and there are some more work needed for that to happen.

ZheyuYe:
I can confirm that there is a bug here as 
`
<<<MXNetError: Check failed: assign(&dattr, vec.at(i)): Incompatible attr in node _npi_stack at 1-th input: expected int32, got int64
`

PR #1285, opened 2020-08-02
[CI] Try to fix CI

sxjscience:
CI is reporting the `libprotobuf` not found error in MacOS. Try to fix.


leezu:
It's a regression in mxnet pip wheel introduced by

PR #1288, opened 2020-08-06
Update Dockerfile

ZheyuYe:
## Description ##
As titile

### DIFF
--- a/tools/batch/docker/Dockerfile
+++ b/tools/batch/docker/Dockerfile
@@ -23,9 +23,6 @@ FROM nvidia/cuda:10.2-cudnn7-devel-ubuntu18.04
 
  RUN pip3 install --upgrade pip && pip3 install awscli && pip3 install --pre 'mxnet-cu102' -f https://dist.mxnet.io/python
 
- RUN ln -s /usr/bin/python3 /usr/bin/python && \
-    ln -s /usr/bin/pip3 /usr/bin/pip
-
  # Avoid ascii errors when reading files in Python
  RUN apt-get install -y \
    locales && \
--- a/tools/batch/docker/Dockerfile
+++ b/tools/batch/docker/Dockerfile
@@ -22,6 +22,16 @@ FROM nvidia/cuda:10.2-cudnn7-devel-ubuntu18.04
       rm -rf /var/lib/apt/lists/*
 
  RUN pip3 install --upgrade pip && pip3 install awscli && pip3 install --pre 'mxnet-cu102' -f https://dist.mxnet.io/python
+
+ RUN ln -s /usr/bin/python3 /usr/bin/python && \
+    ln -s /usr/bin/pip3 /usr/bin/pip
+
+ # Avoid ascii errors when reading files in Python
+ RUN apt-get install -y \
+   locales && \
+   locale-gen en_US.UTF-8
+ ENV LANG='en_US.UTF-8' LANGUAGE='en_US:en' LC_ALL='en_US.UTF-8'
+
  RUN git clone https://github.com/dmlc/gluon-nlp
  WORKDIR gluon-nlp
  ADD gluon_nlp_job.sh .


tools/batch/docker/Dockerfile
```diff
@@ -22,6 +22,16 @@ FROM nvidia/cuda:10.2-cudnn7-devel-ubuntu18.04
       rm -rf /var/lib/apt/lists/*
 
  RUN pip3 install --upgrade pip && pip3 install awscli && pip3 install --pre 'mxnet-cu102' -f https://dist.mxnet.io/python
+
+ RUN ln -s /usr/bin/python3 /usr/bin/python && \
+    ln -s /usr/bin/pip3 /usr/bin/pip
```
sxjscience, line 27: Should not do with `ln`.

src/gluonnlp/layers.py
```diff
@@ -356,8 +356,7 @@ def __init__(self, mode='erf'):
 
     def hybrid_forward(self, F, x):
         if self._mode == 'erf':
-            # TODO Investigate the precision of F.npx.leaky_relu(x, act_type='gelu')
-            return x * 0.5 * (1.0 + F.npx.erf(x / math.sqrt(2.0)))
+            return F.npx.leaky_relu(x, act_type='gelu')
```
sxjscience, line 359: Why does it appear again?

sxjscience:
@ZheyuYe How about not to change these system configurations? Instead, we can revise the conversion scripts.

tools/batch/docker/Dockerfile
sxjscience: I think you may try to use `update-alternatives`.

ZheyuYe:
@leezu I am setting en_US.UTF-8 as default to fix the file openning issue in python3.

ZheyuYe:
@sxjscience Reviseing our scripts and replacing `python` with `python3` would be a good idea

sxjscience:
@ZheyuYe Actually, you may use update-alternatives, which is the standard way in Ubuntu to convert python --> python3.

tools/batch/docker/Dockerfile
ZheyuYe: like `update-alternatives sudo update-alternatives --install /usr/bin/python python /usr/bin/python3` ?

leezu: It may be better to revise the script to use `python3` as it improves portability

sxjscience:
@leezu Do you know if python3 still point to the write python if we use conda?

sxjscience:
@leezu @ZheyuYe I checked the dockerfile in official tensorflow:
https://github.com/tensorflow/tensorflow/blob/1e6fa32dfe9d3f1631ed6218985426d3f6ca4860/tensorflow/tools/dockerfiles/dockerfiles/devel-gpu-jupyter.Dockerfile#L111

They are also using `ln -s` to update the python to python3. I think we can do the same?

tools/batch/docker/Dockerfile
sxjscience: After checking the TF dockerfile  https://github.com/tensorflow/tensorflow/blob/1e6fa32dfe9d3f1631ed6218985426d3f6ca4860/tensorflow/tools/dockerfiles/dockerfiles/devel-gpu-jupyter.Dockerfile#L111 , the correct approach might be to use `ln -s` but create the link in `/usr/local/bin/python`.

leezu:
> @leezu I am setting en_US.UTF-8 as default to fix the file openning issue in python3.

It should already be the default. Can you provide more information on the error and why your change is required? Thanks

@sxjscience why introduce a divergence between user system and our docker containers? It may be simpler to follow the defaults for both

sxjscience:
@leezu I find that TF also supports sets the ENV: https://github.com/tensorflow/tensorflow/blob/1e6fa32dfe9d3f1631ed6218985426d3f6ca4860/tensorflow/tools/dockerfiles/dockerfiles/devel-gpu-jupyter.Dockerfile#L100

leezu:
> @leezu I find that TF also sets the ENV: https://github.com/tensorflow/tensorflow/blob/1e6fa32dfe9d3f1631ed6218985426d3f6ca4860/tensorflow/tools/dockerfiles/dockerfiles/devel-gpu-jupyter.Dockerfile#L100

That's related to a bug fixed in Python 3.3 and 3.4, but there may be other issues. @ZheyuYe can you provide more information on the issue you experience?

I checked the default settings of the nvidia docker and indeed no language is configured. However, we use the same settings for the MXNet CI without problem, so we should at least document why the discrepancy is required here.

```
root@b20f04e442df:/# locale
LANG=
LANGUAGE=
LC_CTYPE="POSIX"
LC_NUMERIC="POSIX"
LC_TIME="POSIX"
LC_COLLATE="POSIX"
LC_MONETARY="POSIX"
LC_MESSAGES="POSIX"
LC_PAPER="POSIX"
LC_NAME="POSIX"
LC_ADDRESS="POSIX"
LC_TELEPHONE="POSIX"
LC_MEASUREMENT="POSIX"
LC_IDENTIFICATION="POSIX"
LC_ALL=
```

ZheyuYe:
@leezu The issue I encountered occurred when local `vocab.txt file` was opened by `python3` in 
https://github.com/dmlc/gluon-nlp/blob/9e268c0e5e0393e4ce19a4161ef2a2f8eb56797d/scripts/conversion_toolkits/convert_electra.py#L117

Error message:
```
File "convert_electra.py", line 439, in <module>
args.electra_path)
File "convert_electra.py", line 200, in convert_tf_model
cfg, vocab_path = convert_tf_assets(model_dir, model_size, electra_path)
File "convert_electra.py", line 117, in convert_tf_assets
vocab_size = len(open(vocab_path, 'rU').readlines())
File "/usr/lib/python3.6/encodings/ascii.py", line 26, in decode
return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 3793: ordinal not in range(128)
```
Reproduced command:
```
python3 tools/batch/submit-job.py \
--region us-east-1 \
--source-ref conversion_tf2 \ 
--job-type g4dn.4x \
--name test_convert_electra \
--work-dir scripts/conversion_toolkits \
--remote https://github.com/zheyuye/gluon-nlp/ \
--command 'python3 convert_electra.sh | tee stdout.log' \
--wait 
```
See the logs in AWS Batch for more details.

leezu:
@ZheyuYe you can't assume that the system encoding is `utf-8`. A better approach would be to specify `encoding='utf-8'` in https://github.com/dmlc/gluon-nlp/blob/9e268c0e5e0393e4ce19a4161ef2a2f8eb56797d/scripts/conversion_toolkits/convert_electra.py#L117

Also note that the `U` for universal newlines is deprecated and has no effect in Python 3.

ZheyuYe:
I am closing this PR and creating an another PR to specify the encoding type

PR #1291, opened 2020-08-07
Fix electra 

ZheyuYe:
## Description ##
Fix electra pretraining script and data preprocessing script

## Comments ##
cc @sxjscience


sxjscience:
Need to test for the data_preprocessing.

sxjscience:
I'll approve + merge this first but we should add the test later.

PR #1292, opened 2020-08-10
[Numpy] Benchmark the backbone models + Some fixes + Always use python3 + Fix conversion tool

sxjscience:
## Description ##
After this PR is merged, you can
- [x] Benchmarking the performance of backbone models in GluonNLP by running `scripts/benchmarks/benchmark_gluonnlp.py`
- [x] Always use `python3` and `pip3` as discussed with @szha and @leezu .
- [x] ~~Test for the conversion scripts.~~ (Next PR, this one will revise the conversion scripts)

cc @dmlc/gluon-nlp-team


scripts/pretraining/pretraining_utils.py
```diff
@@ -56,9 +56,8 @@ def tokenize_lines_to_ids(lines, tokenizer):
     return results
 
 
-def get_all_features(x):
-    """
-    Get the feature data in numpy form.
+def get_all_features(file_list, output_file, tokenizer, max_seq_length, short_seq_prob):
+    """Get the feature data in numpy form.
```
ZheyuYe, line 60: This function is used with `pool.imap`, and this modification will raise an error with message `TypeError: get_all_features() missing 4 required positional arguments: 'output_file', 'tokenizer', 'max_seq_length', and 'short_seq_prob'`
 https://github.com/dmlc/gluon-nlp/blob/9e268c0e5e0393e4ce19a4161ef2a2f8eb56797d/scripts/pretraining/data_preprocessing.py#L74

BTW, it is also used in https://github.com/dmlc/gluon-nlp/blob/9e268c0e5e0393e4ce19a4161ef2a2f8eb56797d/scripts/pretraining/pretraining_utils.py#L273-L274

scripts/pretraining/pretraining_utils.py
```diff
@@ -190,7 +194,7 @@ def convert_to_npz(all_features, output_file=None):
         npz_outputs['valid_lengths'] = np.array(valid_lengths, dtype='int32')
         np.savez_compressed(output_file, **npz_outputs)
         logging.info("Saved {} features in {} ".format(len(all_features), output_file))
-    return (input_ids, segment_ids, valid_lengths)
+    return input_ids, segment_ids, valid_lengths
```
ZheyuYe, line 197: The return conflicts with the call https://github.com/dmlc/gluon-nlp/blob/9e268c0e5e0393e4ce19a4161ef2a2f8eb56797d/scripts/pretraining/pretraining_utils.py#L86

sxjscience: I think there is no need to surround it with a tuple and return. Basically, adding tuple/not is equivalent.

ZheyuYe: Yeah that is true

scripts/pretraining/pretraining_utils.py
sxjscience: Fixed

sxjscience:
Initial numbers obtained via the benchmark:

See the shared google sheet: https://docs.google.com/spreadsheets/d/1ohQ2P4PNDxciz1_gWJaGzD-LeXsvX5MSe-qMoqnRaxc/edit?usp=sharing

PR #1301, opened 2020-08-17
[FEATURE]Horovod support for training transformer (PART 2)

hymzoque:
### Essentials ###
- [ ] Code is well-documented

### Changes ###
-  add horovodrun example in machine_translation/README.md 
-  transformer base/large result (test and valid bleu)
-  ShardedIterator (to split sampler) and testcase
-  add test case to test situation that fail downloading in test_util_misc.py


scripts/machine_translation/README.md
```diff
@@ -30,30 +30,38 @@ python3 train_transformer.py \
...
+    --end 39 \
+    --save-path transformer_base_wmt2014_en_de_${SUBWORD_ALGO}/epoch_avg_30_39.params
 ```
 
-
 Use the following command to inference/evaluate the Transformer model:
 
 ```bash
 SUBWORD_MODEL=yttm
-python3 evaluate_transformer.py \
```
ZheyuYe, line 55: plz use python3 so that it works on any systems

scripts/datasets/machine_translation/wmt2014_ende.sh
```diff
@@ -23,7 +23,7 @@ nlp_preprocess clean_tok_para_corpus --src-lang ${SRC} \
                       --src-corpus train.raw.${SRC} \
                       --tgt-corpus train.raw.${TGT} \
                       --min-num-words 1 \
-                      --max-num-words 100 \
+                      --max-num-words 250 \
```
sxjscience, line 26: I think 250 is the truncation length after BPE tokenization. Have you verified that we should truncate to 250 before the BPE tokenization?

scripts/machine_translation/README.md
```diff
@@ -97,7 +105,7 @@ python3 train_transformer.py \
     --gpus 0,1,2,3
 ```
 
-Use the average_checkpoint cli to average the last 10 checkpoints
+Use the average_checkpoint cli to average the last 10 update checkpoints
```
sxjscience, line 108: Just use `average the last 10 checkpoints`.

tests/test_data_sampler.py
```diff
@@ -137,13 +137,18 @@ def test_split_sampler_even_size(num_samples, num_parts):
...
+    samplers = []
+    for index in range(num_parts):
+        samplers.append(s.BoundedBudgetSampler(seq_lengths, max_num_tokens, max_num_sentences,
+                                               required_batch_size_multiple, shuffle, seed,
+                                               num_parts, index))
     total_sampled_ids = []
-    for batch_sample_ids in sampler:
-        total_sampled_ids.extend(batch_sample_ids)
+    for sampler in samplers:
+        print(sampler)
```
ZheyuYe, line 150: Should we avoid printing in test?

scripts/machine_translation/train_transformer.py
```diff
@@ -328,9 +333,9 @@ def train(args):
     cfg.freeze()
     model = TransformerModel.from_cfg(cfg)
     model.initialize(mx.init.Xavier(magnitude=args.magnitude),
-                     ctx=ctx_l)
+                     ctx=args.ctx_l)
```
sxjscience, line 336: I think you should use `ctx_l`.

scripts/machine_translation/train_transformer.py
```diff
@@ -457,7 +461,7 @@ def train(args):
                 sample_data_l = next(train_multi_data_loader)
             except StopIteration:
                 is_last_batch = True
-            if local_rank == 0 and num_params is None:
+            if args.local_rank == 0 and num_params is None:
```
sxjscience, line 464: Why use args.local_rank instead of local_rank?

scripts/machine_translation/train_transformer.py
```diff
@@ -475,35 +479,39 @@ def train(args):
...
                    (n_epoch_train_iters % args.log_interval == 0 or is_last_batch):
                     log_end_time = time.time()
                     log_wc = log_wc.asnumpy()
                     wps = log_wc / (log_end_time - log_start_time)
                     log_avg_loss = (log_avg_loss / log_loss_denom).asnumpy()
                     logging.info('[Epoch {} Batch {}/{}] loss={:.4f}, ppl={:.4f}, '
                                  'throughput={:.2f}K wps, wc={:.2f}K, LR={}'
-                                 .format(epoch_id, processed_batch_num * num_parts, len(train_data_loader),
-                                         log_avg_loss, np.exp(log_avg_loss),
+                                 .format(epoch_id, min(processed_batch_num * args.num_parts, len(train_data_loader)),
```
sxjscience, line 490: Why do we need the min?

src/gluonnlp/data/sampler.py
```diff
@@ -289,12 +288,16 @@ class BoundedBudgetSampler(BaseSampler):
...
+        If the number of batches is not even across all partitions, sample a few extra batches
+        for the ones with fewer batches.
     """
     def __init__(self, lengths: Union[Sequence[int], Sequence[Sequence[int]]],
                  max_num_tokens: int = -1, max_num_sentences: int = -1,
                  required_batch_size_multiple: int = 1,
                  shuffle: bool = True, seed: Optional[int] = None,
-                 num_parts: int = 1, part_index: int = 0):
+                 num_parts: int = 1, part_index: int = 0,
+                 even_size: bool = False):
```
sxjscience, line 300: I suggest wrapping the handling of `even_size` and `num_parts` into another function.

sxjscience: Is it possible to write something similar to https://github.com/NVIDIA/DeepLearningExamples/blob/d875531dd69de92c8424e2254df6f393088e7ef4/PyTorch/Translation/Transformer/fairseq/data/data_utils.py#L65-L89?

scripts/datasets/machine_translation/wmt2014_ende.sh
hymzoque: 250 is consistent with  https://github.com/NVIDIA/DeepLearningExamples/blob/master/PyTorch/Translation/Transformer/examples/translation/prepare-wmt14en2de.sh#L181-L182

scripts/machine_translation/train_transformer.py
hymzoque: `processed_batch_num * args.num_parts > len(train_data_loader)` may happens

tests/test_data_sampler.py
hymzoque: idk, it's consistent with https://github.com/dmlc/gluon-nlp/blob/master/tests/test_data_sampler.py#L37

scripts/machine_translation/train_transformer.py
sxjscience: I think for the purpose of logging, we should just use `processed_batch_num * args.num_parts`.

sxjscience:
For the diff check failure, it might be caused by the fact that the exception is not triggered:
https://github.com/dmlc/gluon-nlp/blob/32e87d4d4aa20a6eb658ee90d765ccffbd160571/src/gluonnlp/utils/misc.py#L562-L568

Thus, we may try to add a test-case which downloads an non-existing file and use `pytest.assertRaises`.

tests/test_data_sampler.py
sxjscience: It's not necessary to `print(sampler)`. However, it does not introduce any bugs because the output stream will be ignored.

scripts/machine_translation/train_transformer.py
hymzoque: ok

tests/test_data_sampler.py
hymzoque: ok

scripts/datasets/machine_translation/wmt2014_ende.sh
hymzoque: it should be 100

scripts/machine_translation/README.md
```diff
@@ -30,30 +30,38 @@ python3 train_transformer.py \
...
     --gpus 0,1,2,3
 ```
 
-Use the average_checkpoint cli to average the last 10 checkpoints
+Or training via horovod
+```
+mpirun -np 4 -H localhost:4 python3 train_transformer.py \
+    --comm_backend horovod \
+    ...
+```
```
sxjscience, line 48: Do you have any special reason to use mpirun instead of horovodrun? @ZheyuYe

ZheyuYe: At first I cant run the scripts via 'horovodrun' and switched to 'mpirun'，probably because I didn't install it properly. Let me re-try 'horovodrun'.

src/gluonnlp/data/sampler.py
hymzoque: changed to ShardedIterator class in sampler.py

scripts/machine_translation/README.md
hymzoque: changed mpirun to horovodrun

sxjscience:
I find that the codecov is not very stable, especially the diff hit. @szha @leezu Would you know how could we improve it?

szha:
assuming no race issue in result upload, this can be a result of non-determinism in tests, and the code path in the same test can vary.

if that's the case, we can mock the randomness to deterministically test each path. or if we still require the randomness, we can increase the trial number

scripts/machine_translation/README.md
```diff
@@ -30,30 +30,37 @@ python3 train_transformer.py \
...
     --warmup_init_lr 0.0 \
     --seed 123 \
     --gpus 0,1,2,3
 ```
 
+Or training via horovod
+```
+horovodrun -np 4 -H localhost:4 python3 train_transformer.py \
+    --comm_backend horovod \
+    ...
```
sxjscience, line 47: Rather than `...`, would you paste the whole command?

PR #1314, opened 2020-08-25
[DEV] automatically generate date suffix for dev versions

szha:
## Description ##
automatically generate date suffix for dev versions

### Essentials ###
- [x] Code is well-documented

### Changes ###
- [x] automatically generate date suffix for dev versions

cc @dmlc/gluon-nlp-team

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+from datetime import datetime
 import io
 import os
 import re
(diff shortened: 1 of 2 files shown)


szha:
with this change, I think we won't really need a nightly build as people can install from github directly and still has a version-number/date to reference.

sxjscience:
In AutoGluon, the logic is to always use the dev version if it’s not a release: https://github.com/awslabs/autogluon/blob/30169a32ece713b676a77c785a2b6e6f2644bb80/setup.py#L17. Should we adopt the same logic?

szha:
I don't think relying on an environment variable is the way to go. For one, this creates a leaky abstraction outside of version control. We have a proper path for managing different versions due to our use of release branches and tags, so there really isn't a need for the autogluon approach for us.

Issue #1315, opened 2020-08-25
Bugs for run_electra.py at master branch


ZiyueHuang:
### Bug for Horovod support

At https://github.com/dmlc/gluon-nlp/blob/master/scripts/pretraining/run_electra.py#L418, `num_samples_per_update` depends on the data batch, and thus could be different across the horovod processes (note that Horovod will launch a process for each GPU card), meaning that the parameters maintained in each horovod process (since we use `trainer.allreduce_grads()`) might diverge.

We can normalize the loss by batch_size on each worker before `trainer.allreduce_grads()`, to avoid syncing `num_samples_per_update` across all workers.

### Bug for MLM loss

In the current implementation, the MLM loss actually doesn't take into account `masked_weights`.

At https://github.com/dmlc/gluon-nlp/blob/master/scripts/pretraining/run_electra.py#L383-L384, `masked_weights.reshape(-1)` should be replaced by `mx.np.expand_dims(masked_weights, axis=-1)`, please see the doc of `SoftmaxCELoss` at https://github.com/apache/incubator-mxnet/pull/19010/files or the example usage in bert at https://github.com/dmlc/gluon-nlp/blob/v0.10.x/scripts/bert/pretraining_utils.py#L416-L418

Note that

`mlm_scores`'s shape: (batch_size, num_masked_positions, vocab_size)
`unmasked_tokens`'s shape: (batch_size, num_masked_positions)
`masked_weights`'s shape: (batch_size, num_masked_positions)

If we comment out `mlm_loss_fn.hybridize()` and add two lines before/after https://github.com/apache/incubator-mxnet/blob/master/python/mxnet/gluon/loss.py#L408: `print('loss.shape before weighting ', loss.shape, ' sample_weight.shape ', sample_weight.shape)` and `print('loss.shape after weighting ', loss.shape)`, running the current script run_electra.py we will get

```
loss.shape before weighting  (64, 19, 1)  sample_weight.shape  (1216,)
loss.shape after weighting  (64, 19, 1216)
```
note that 1216 = 64 x 19

cc @sxjscience @ZheyuYe 

### Error Message
(Paste the complete error message, including stack trace.)

## To Reproduce
(If you developed your own code, please provide a short script that reproduces the error. For existing examples, please provide link.)

### Steps to reproduce
(Paste the commands you ran that produced the error.)

1.
2.

## What have you tried to solve it?

1.
2.

## Environment

[...]

PR #1319, opened 2020-08-26
fix https://github.com/dmlc/gluon-nlp/issues/1315

ZiyueHuang:
## Description ##
For each worker, we should normalize the loss by the number of elements to improve numerical stability and can avoid setting `step_size` in trainer.update(), and this is consistent with the official TF implementation, see https://github.com/google-research/electra/blob/master/run_pretraining.py#L181 and https://github.com/google-research/electra/blob/master/run_pretraining.py#L206.

To check the correctness of the Horovod support, let k denote the number of workers,

Ground truth:
```
g = \frac{1}{k} \sum_{i=1}{k} g_i

update_gt = clip_grad_global_norm(g, max_norm)
```

In our code
```
G = \sum_{i=1}{k} g_i     (by trainer.allreduce_grads())

update = \frac{1}{k} * clip_grad_global_norm(G, max_norm * k)

         (note that \frac{1}{k} is the scaling factor inside hvd.trainer)

       = \frac{1}{k} * clip_grad_global_norm(g * k, max_norm * k)

       = \frac{1}{k} * k * clip_grad_global_norm(g, max_norm)

       = update_gt
```
And this part of implementation for Horovod support is consistent with BERT script, see https://github.com/dmlc/gluon-nlp/blob/v0.10.x/scripts/bert/run_pretraining.py.

@sxjscience  Could you also check the correctness?

Also, we need re-train the model and adjust the hyper-parameters accordingly.

### Essentials ###
- [ ] Code is well-documented

### Changes ###

## Comments ##

cc @dmlc/gluon-nlp-team


scripts/pretraining/run_electra.py
```diff
@@ -378,20 +377,22 @@ def train(args):
                 with mx.autograd.record():
                     mlm_scores, rtd_scores, corrupted_tokens, labels = model(
                         masked_input_ids, segment_ids, valid_lengths, unmasked_tokens, masked_positions)
-                    # the official implementation takes the sum of each batch inside the loss function
-                    # while SigmoidBinaryCrossEntropyLoss and SoftmaxCELoss takes the mean value
+                    num_elements = masked_weights.sum()
                     mlm_loss = mlm_loss_fn(
-                        mlm_scores, unmasked_tokens, masked_weights.reshape(-1)).mean() / (masked_weights.mean() + 1e-6)
+                        mx.nd.reshape(mlm_scores.as_nd_ndarray(), (-3, -1)).as_np_ndarray(),
```
sxjscience, line 382: For advanced reshape, you can use `mx.npx.reshape(mlm_scores, (-4, -2))`

sxjscience:
LGTM. Need to revise the usage of `nd.reshape` --> `npx.reshape`.

scripts/pretraining/run_electra.py
```diff
@@ -402,24 +403,28 @@ def train(args):
             log_rtd_loss += sum([ele.as_in_ctx(ctx_l[0])
                                  for ele in rtd_loss_l]).asnumpy()
             log_total_loss += sum([ele.as_in_ctx(ctx_l[0])
-                                   for ele in loss_l]).asnumpy() * loss_denom
+                                   for ele in loss_l]).asnumpy()
```
ZheyuYe, line 406: Does `log_total_loss` need to multiply by this coefficient `args.num_accumulated` here?

sxjscience: I think each loss was divided by `args.num_accumulated` so there is no need to multiply `args.num_accumulated`

ZiyueHuang: @sxjscience `log_total_loss` is only used in logging.info(). And I notice that the logging phase also has some problems (but irrelevant to the training), I will fix this later.

ZheyuYe: During the log recording phase, is `log_total_loss` equal to the weighted sum of `log_rtd_loss` and `log_mlm_loss` for now?

ZiyueHuang:
I've made some change. Could you please review again?

PR #1334, opened 2020-09-01
[WIP] distributed training

ZiyueHuang:
## Description ##
Based on this branch (https://github.com/ZiyueHuang/byteps/tree/mx2), we can perform distributed training for the electra model (and other models). However, there are two issues:

* We should first call `trainer._init_params()` before the first iteration (forward/backward) to synchronize the parameters across all workers, otherwise `trainer` will call `_init_params` inside `allreduce_grads` (see mxnet/python/mxnet/gluon/trainer.py), thus the synchronization of the parameters actually happen after the first forward/backward computation, meaning that in the first iteration the gradients on different workers are computed w.r.t. different parameters. But in practice this may not be a severe problem, as only the first gradient descent step is not totally correct.

When the model in gluon-nlp confirms to the new coding standard (removing defer_init for all parameters @sxjscience ), then we can directly call `trainer._init_params` after `model.initialize`.

Actually I am a little confused about the semantic meaning of defer_init in the numpy mode, because shape with zeros (such as `(0, 5)`) in the numpy mode will be treated as scalar tensors' shape and zero-size tensors' shape, instead of as unknown in the legacy mode.

* I found that we have to build MXNet and BytePS from source on our target machine. Using `pip install mxnet` (I have tried several wheels in August) will not work for BytePS, either core dump immediately after `import bps.mxnet` due to undefined symbols (seems because some symbols are not exported into binary, fixed after 0820 wheel), or mysterious segfault (maybe due to c++ ABI issues or other compiler issues, as the mxnet wheels and BytePS (which is setup on our target machine) may not be compiled using the same version of gcc. Besides, there are no related compiler flags such as D_GLIBCXX_USE_CXX11_ABI for MXNet in BytePS's setup.py).

### Essentials ###
- [ ] Code is well-documented

### Changes ###

## Comments ##

cc @dmlc/gluon-nlp-team

### DIFF
--- a/scripts/pretraining/run_electra.py
+++ b/scripts/pretraining/run_electra.py
@@ -22,6 +22,10 @@ try:
     import horovod.mxnet as hvd
 except ImportError:
     pass
+try:
+    import byteps.mxnet as bps
+except ImportError:
+    pass
 
 mx.npx.set_np()
 
@@ -118,7 +122,7 @@ def parse_args():
                         help='The scale size of the generator layer')
     # Communication
     parser.add_argument('--comm_backend', type=str, default='device',
-                        choices=['horovod', 'dist_sync_device', 'device'],
+                        choices=['byteps', 'horovod', 'dist_sync_device', 'device'],
                         help='Communication backend.')
     parser.add_argument('--gpus', type=str, default='0',
                         help='list of gpus to run, e.g. 0 or 0,2,5. -1 means using cpu.')
@@ -316,6 +320,8 @@ def train(args):
                                  })
     if args.comm_backend == 'horovod':
         trainer = hvd.DistributedTrainer(param_dict, args.optimizer, optimizer_params)
+    elif args.comm_backend == 'byteps':
+        trainer = bps.DistributedTrainer(param_dict, args.optimizer, optimizer_params)
     else:
         trainer = mx.gluon.Trainer(param_dict, args.optimizer, optimizer_params,
                                    update_on_kvstore=False)
@@ -414,8 +420,8 @@ def train(args):
         total_norm, ratio, is_finite = clip_grad_global_norm(
             params, args.max_grad_norm * num_workers)
 
-        if args.comm_backend == 'horovod':
-            # Note that horovod.trainer._scale is default to num_workers,
+        if args.comm_backend == 'horovod' or args.comm_backend == 'byteps':
+            # Note that hvd.trainer._scale and bps.trainer._scale are default to num_workers,
             # thus trainer.update(1) will scale the gradients by 1./num_workers
             trainer.update(1, ignore_stale_grad=True)
         else:
--- a/src/gluonnlp/utils/misc.py
+++ b/src/gluonnlp/utils/misc.py
@@ -637,6 +637,20 @@ def init_comm(backend, gpus):
         is_master_node = rank == local_rank
         ctx_l = [mx.gpu(local_rank)]
         logging.info('GPU communication supported by horovod')
+    elif backend == 'byteps':
+        try:
+            import byteps.mxnet as bps
+        except ImportError:
+            logging.info('BytePS must be installed.')
+            sys.exit(1)
+        bps.init()
+        store = None
+        num_workers = bps.size()
+        rank = bps.rank()
+        local_rank = bps.local_rank()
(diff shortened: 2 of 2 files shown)


ZheyuYe (commented):
Looks good and build MXNet from source will work for distributed training

ZiyueHuang:
Just found that the API behavior in BytePS master branch changes recently...... Not sure if that is intended or bugs. Tracked here (https://github.com/bytedance/byteps/issues/292).

sxjscience:
@ZiyueHuang  You may merge the upstream/master since we recently fixed the CI.

szha:
> undefined symbols

What are they?

szhengac:
Horovod also has similar issue that the actual broadcast happens after the first iteration. See [here](https://github.com/horovod/horovod/blob/9464e20ab2565f25105d5d7f6194ca90d6fca9e5/horovod/mxnet/__init__.py#L154-L158)

sxjscience:
To avoid the problems of horovod, in GluonNLP v1, the convention is to not rely on deferred initialization when implementing the model. For example, for dense layer, we should always give the in_units.

ZiyueHuang:
Sorry for the late reply.

@szha  core dump due to undefined symbols is fixed after 0820 wheel, and I didn't record the undefined symbols. For the segfault, below is the stack trace
```
Fatal Error: Segmentation fault
Stack trace:
Segmentation fault (core dumped)
Traceback (most recent call last):
  File "/home/ubuntu/anaconda3/bin/bpslaunch", line 4, in <module>
    __import__('pkg_resources').run_script('byteps==0.2.4', 'bpslaunch')
  File "/home/ubuntu/anaconda3/lib/python3.7/site-packages/pkg_resources/__init__.py", line 667, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/home/ubuntu/anaconda3/lib/python3.7/site-packages/pkg_resources/__init__.py", line 1464, in run_script
    exec(code, namespace, namespace)
  File "/home/ubuntu/anaconda3/lib/python3.7/site-packages/byteps-0.2.4-py3.7-linux-x86_64.egg/EGG-INFO/scripts/bpslaunch", line 213, in <module>
    launch_bps()
  File "/home/ubuntu/anaconda3/lib/python3.7/site-packages/byteps-0.2.4-py3.7-linux-x86_64.egg/EGG-INFO/scripts/bpslaunch", line 206, in launch_bps
    t[i].join()
  File "/home/ubuntu/anaconda3/lib/python3.7/site-packages/byteps-0.2.4-py3.7-linux-x86_64.egg/EGG-INFO/scripts/bpslaunch", line 34, in join
    raise self.exc
  File "/home/ubuntu/anaconda3/lib/python3.7/site-packages/byteps-0.2.4-py3.7-linux-x86_64.egg/EGG-INFO/scripts/bpslaunch", line 27, in run
    self.ret = self._target(*self._args, **self._kwargs)
  File "/home/ubuntu/anaconda3/lib/python3.7/site-packages/byteps-0.2.4-py3.7-linux-x86_64.egg/EGG-INFO/scripts/bpslaunch", line 177, in worker
    stdout=sys.stdout, stderr=sys.stderr, shell=True)
  File "/home/ubuntu/anaconda3/lib/python3.7/subprocess.py", line 363, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command 'python ./test.py' returned non-zero exit status 139.
```

@sxjscience @szhengac  Horovod doesn't have similar issue, since in `hvd.broadcast_parameters` when there are defer_init parameters, it will register a hook to `param. init_impl` (see https://github.com/horovod/horovod/blob/9464e20ab2565f25105d5d7f6194ca90d6fca9e5/horovod/mxnet/__init__.py#L121-L127), thus immediately after the actual initialization, Horovod will broadcast the parameters, then forward/backward. However, there are no similar mechanisms in BytePS to handle the logic of defer_init.

leezu:
>  Actually I am a little confused about the semantic meaning of defer_init in the numpy mode, because shape with zeros (such as (0, 5)) in the numpy mode will be treated as scalar tensors' shape and zero-size tensors' shape, instead of as unknown in the legacy mode.

You can report it as a bug in MXNet issue tracker

PR #1348, opened 2020-09-03
[Fix][SageMaker] Make sure that the installation works in SageMaker

sxjscience:
## Description ##
Make sure the installation works on SageMaker Notebook with `conda_python3` kernel.

cc @dmlc/gluon-nlp-team

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ requirements = [
     'tokenizers==0.8.1',
     'click>=7.0',  # Dependency of youtokentome
     'youtokentome>=1.0.6',
-    'fasttext==0.9.1'  # Fix to 0.9.1 due to https://github.com/facebookresearch/fastText/issues/1052
+    'fasttext>=0.9.1,!=0.9.2'  # Fix to 0.9.1 due to https://github.com/facebookresearch/fastText/issues/1052
 ]
 
 setup(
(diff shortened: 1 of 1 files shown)


sxjscience:
@szha @leezu @barry-jin @szhengac  Currently I'm fixing the problem manually but we may consider to add some sagemaker-based unittests to make sure that everything works on SageMaker Notebook.

setup.py
```diff
@@ -47,7 +47,7 @@ def find_version(*file_paths):
     'tokenizers==0.8.1',
     'click>=7.0',  # Dependency of youtokentome
     'youtokentome>=1.0.6',
-    'fasttext>=0.9.2'
+    'fasttext==0.9.1'  # Fix to 0.9.1 due to https://github.com/facebookresearch/fastText/issues/1052
```
szha, line 50: can we do `!=` to blacklist some versions instead? my hope is that our CI can catch installation problems as soon as they happen. if you adopt this approach, then we should also specify any minimum version requirement if we have any.

sxjscience: Should we do `fasttext>=0.9.1, !=0.9.2` ? Also, the problem is actually related to the fact that the latest fasttext does not work on SageMaker.

ZheyuYe (approved):

sxjscience:
Thanks @preeyank5 for reporting the issue. This should have been fixed. However, we still need to figure out the way to test on SageMaker Notebook instance.

Issue #1347, opened 2020-09-03
[Bug] [Typo] Fix convert_all.sh


sxjscience:
Fix these two lines

https://github.com/dmlc/gluon-nlp/blob/e0b293e9ba515f8b4c58671232bd7c1b99522ebd/scripts/conversion_toolkits/convert_all.sh#L3-L4

`convert_bert_from_tf_hub.sh` --> `convert_bert.sh`
`convert_albert_from_tf_hub.sh` --> `convert_albert.sh`

ZheyuYe:
Closed by https://github.com/dmlc/gluon-nlp/pull/1350

PR #1350, opened 2020-09-04
remove convert_all.sh

ZheyuYe:
## Description ##
Will fix https://github.com/dmlc/gluon-nlp/issues/1347

cc @sxjscience


sxjscience (approved):

szha:
@sxjscience do we really need this script? I'm not sure if this is useful to users and this file seems like a place where new addition would be missed.

sxjscience:
It's also fine to directly move this file.

sxjscience:
@ZheyuYe Would you remove this file?

sxjscience:
@szha Would you have other concerns?

PR #1353, opened 2020-09-07
Fix the correctness of the Horovod support on squad

ZheyuYe:
## Description ##
Fix squad to support horovod as https://github.com/dmlc/gluon-nlp/pull/1319 did for pretraining electra

Got a result of 77.62/80.39 for mobilebert. See https://github.com/dmlc/gluon-nlp/tree/master/scripts/question_answering for the comparison

## Comments ##
@sxjscience @ZiyueHuang 

cc @dmlc/gluon-nlp-team

### DIFF
--- a/scripts/question_answering/run_squad.py
+++ b/scripts/question_answering/run_squad.py
@@ -6,7 +6,6 @@ Question Answering with Pretrained Language Model
 import os
 import json
 import time
-import random
 import logging
 import argparse
 import functools
@@ -119,6 +118,8 @@ def parse_args():
                         help='Whether to shuffle the train dataset before splitting')
     parser.add_argument('--inside_split_shuffle', action='store_false',
                         help='Whether to shuffle the train dataset inside each splitted shards')
+    parser.add_argument('--pre_shuffle_seed', type=int, default=100,
+                        help='Random seed for pre split shuffle')
     parser.add_argument('--round_to', type=int, default=None,
                         help='The length of padded sequences will be rounded up to be multiple'
                              ' of this argument. When round to is set to 8, training throughput '
@@ -462,7 +463,8 @@ def train(args):
     logging.info('After Chunking, #Train Sample/Is Impossible = {}/{}'
                  .format(len(train_dataset), num_impossible))
     if args.pre_split_shuffle:
-        random.shuffle(train_dataset)
+        rs = np.random.RandomState(args.pre_shuffle_seed)
+        rs.shuffle(train_dataset)
     sampler = SplitSampler(
         len(train_dataset),
         num_parts=num_workers,
--- a/scripts/question_answering/run_squad.py
+++ b/scripts/question_answering/run_squad.py
@@ -419,10 +419,11 @@ def setup_logging(args, local_rank):
     """
     logging_config(args.output_dir,
                    name='finetune_squad{}'.format(args.version),# avoid race
+                   overwrite_handler=True,
                    console=(local_rank == 0))
     logging.info(args)
     set_seed(args.seed)
-    logging.debug('Random seed set to {}'.format(args.seed)
+    logging.debug('Random seed set to {}'.format(args.seed))
 
 def train(args):
     store, num_workers, rank, local_rank, is_master_node, ctx_l = init_comm(
--- a/scripts/question_answering/run_squad.py
+++ b/scripts/question_answering/run_squad.py
@@ -114,10 +114,16 @@ def parse_args():
     parser.add_argument('--max_query_length', type=int, default=64,
                         help='The maximum number of tokens for the query. Questions longer than '
                              'this will be truncated to this length. default is 64')
-    parser.add_argument('--pre_split_shuffle', action='store_false',
-                        help='Whether to shuffle the train dataset before splitting')
-    parser.add_argument('--inside_split_shuffle', action='store_false',
-                        help='Whether to shuffle the train dataset inside each splitted shards')
+    parser.add_argument("--do_pre_split_shuffle", dest='pre_split_shuffle',
+                        action="store_true", help="Whether to shuffle the train dataset before splitting")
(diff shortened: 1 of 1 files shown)


scripts/question_answering/run_squad.py
```diff
@@ -564,7 +571,7 @@ def train(args):
                     sel_answerable_logits = answerable_logits[batch_idx, is_impossible]
                     span_loss = - 0.5 * (sel_start_logits + sel_end_logits).sum()
                     answerable_loss = -0.5 * sel_answerable_logits.sum()
```
sxjscience, line 573: If that's the case, you will need to use `mean()`

scripts/question_answering/run_squad.py
```diff
@@ -574,27 +581,27 @@ def train(args):
...
-            # \sum_{n=1}^N g_n / loss_denom
-            # Thus, in order to clip the average gradient
-            #   \frac{1}{N} \sum_{n=1}^N      -->  clip to args.max_grad_norm
-            # We need to change the ratio to be
-            #  \sum_{n=1}^N g_n / loss_denom  -->  clip to args.max_grad_norm  * N / loss_denom
             total_norm, ratio, is_finite = clip_grad_global_norm(
-                params, args.max_grad_norm * num_samples_per_update / loss_denom)
+                params, args.max_grad_norm * num_workers)
         else:
             total_norm = grad_global_norm(params)
```
sxjscience, line 594: We will need to divide by `num_workers` when reporting the total_norm.

ZheyuYe:
Please do not merge this till I re-run the mobilebert with horovod again

ZheyuYe:
@sxjscience  I got the following results after the revision
```
{
    "exact": 41.20272888065358,
    "f1": 44.673086401387295,
    "total": 11873,
    "HasAns_exact": 82.52361673414305,
    "HasAns_f1": 89.47428388051136,
    "HasAns_total": 5928,
    "NoAns_exact": 0.0,
    "NoAns_f1": 0.0,
    "NoAns_total": 5945,
    "best_exact": 77.25090541564896,
    "best_exact_thresh": -1.8302667140960693,
    "best_f1": 80.05376875130503,
    "best_f1_thresh": -1.6563208103179932,
    "best_ckpt": "google_uncased_mobilebert_squad2.0_20615.params"
}
```
which is slightly lower than previous results (**80.54/77.81**) that fine-tuned without horovod.

However, horovod do speed up the fine-tuning phrase saving roughly half the time (**2.55 vs 5.05 hours**)

sxjscience:
@ZiyueHuang @szhengac Would you think that the change here is reasonable? The change is that each work will calculate the average loss and then we do allreduce. One issue is that the batch-size of each worker may not be the same. However, I think it should be negligible.

szhengac:
then we should check how large the difference is for the batch size. If it
is large, consider changing sampling strategy to make them closer

sxjscience:
@szhengac @ZheyuYe I just noticed that our current implementation ensures that the batch_size will always be the same.

ZiyueHuang:
`SplitSampler` serves for data sharding among the workers, which may improve the performance of distributed training and is a common practice.

I took a look over `sampler.py` and `dataloader.py`, in the current implementation, for a single worker : it will shuffle the dataset and touch the instances one-by-one. For multiple workers : it will first partition the dataset based on the indexes, and then shuffle within each shard. These two procedures are not equivalent, and the second one may bring a little bias. However, I think this subtle difference may have little influence in practice.

sxjscience:
@ZheyuYe Per-offline discussion with @ZiyueHuang , we find that it may be due to the fact that the current implementation has not shuffled the input samples before sharding:
https://github.com/dmlc/gluon-nlp/blob/3fbe9619d9e68bc665f73c8cdf683213c6edd4d6/src/gluonnlp/data/sampler.py#L551-L564

Basically, we will need to first permutate the training samples before dividing them into shards. For example, it may happen that the initial distribution is highly biased, e.g., the first half are all positive samples and the second half are all negative samples.

Thus, we should shuffle the training indices with the same `pre_sharding_seed` and then shuffle the samples within each shard with the `seed`.

scripts/question_answering/run_squad.py
```diff
@@ -114,6 +114,18 @@ def parse_args():
     parser.add_argument('--max_query_length', type=int, default=64,
                         help='The maximum number of tokens for the query. Questions longer than '
                              'this will be truncated to this length. default is 64')
+    parser.add_argument("--do_pre_split_shuffle", dest='pre_split_shuffle',
+                        action="store_true", help="Whether to shuffle the train dataset before splitting")
+    parser.add_argument("--no_pre_split_shuffle", dest='pre_split_shuffle',
+                        action='store_false', help="Don't lower case input text.")
+    parser.set_defaults(pre_split_shuffle=True)
```
sxjscience, line 121: We may remove these flags and always use pre_split shuffle

sxjscience (approved):
LGTM. One minor comment about the arguments

scripts/question_answering/run_squad.py
```diff
@@ -114,11 +114,6 @@ def parse_args():
...
                         help='The maximum number of tokens for the query. Questions longer than '
                              'this will be truncated to this length. default is 64')
-    parser.add_argument("--do_pre_split_shuffle", dest='pre_split_shuffle',
-                        action="store_true", help="Whether to shuffle the train dataset before splitting")
-    parser.add_argument("--no_pre_split_shuffle", dest='pre_split_shuffle',
-                        action='store_false', help="Don't lower case input text.")
-    parser.set_defaults(pre_split_shuffle=True)
     parser.add_argument("--do_inside_split_shuffle", dest='inside_split_shuffle',
                         action="store_true", help="Whether to shuffle the train dataset before splitting")
     parser.add_argument("--no_inside_split_shuffle", dest='inside_split_shuffle',
```
sxjscience, line 119: Also the inside split shuffle

PR #1390, opened 2020-10-19
[TVM] Add TVM Support

sxjscience:
## Description ##

Add TVM test case + profiling after https://github.com/apache/incubator-tvm/pull/6699 is merged.

- [x] Test case
- [x] Profile

### Essentials ###
- [x] Code is well-documented

cc @dmlc/gluon-nlp-team


.github/workflows/unittests.yml
```diff
@@ -26,11 +26,25 @@ jobs:
...
+            ${{ runner.os }}-ccache
+
+      # Install Linux specific dependencies
       - name: Install Linux dependencies
         if: matrix.os == 'ubuntu-latest'
         # TODO https://github.com/apache/incubator-mxnet/issues/18293
-        run: sudo apt-get install libopenblas-dev
+        run: |
+          sudo apt-get install -y libopenblas-dev ninja-build libedit-dev libxml2-dev
+          sudo bash -c "$(wget -O - https://apt.llvm.org/llvm.sh)"
```
leezu, line 47: Unused?

sxjscience: Do you mean llvm or `libedit-dev libxml2-dev`?

sxjscience: For LLVM, it's used in TVM.

sxjscience: May be it's too slow to ask Github actions to test TVM and I'm considering to remove the check.

sxjscience: I've revised the test case to only test one model when it's on CPU.

sxjscience: I've marked it as resolved. Feel free to update your concern.

src/gluonnlp/utils/misc.py
```diff
@@ -671,3 +673,47 @@ def get_mxnet_visible_ctx():
...
+               'required_pass': ["FastMath"]},
+        'c5': {'target': 'llvm -mcpu=skylake-avx512 -libs=cblas',
+               'use_gpu': False,
+               'opt_level': 3,
+               'required_pass': ["FastMath"]},
+        'p3': {'target': 'cuda -model=v100 -libs=cublas,cudnn',
+               'use_gpu': True,
+               'opt_level': 3,
+               'required_pass': ["FastMath"]}
+    }
```
sxjscience, line 718: @icemelon9 These are the compilation options that I used in GluonNLP.

icemelon9: These configs look good to me.

sxjscience:
Currently blocked by https://github.com/apache/incubator-tvm/pull/6715. Need to fix TVM first since we need to turn on CUBLAS support in GPU by default.

sxjscience:
@szha @leezu @szhengac @barry-jin Would you help review this PR?

.github/workflows/unittests.yml
```diff
@@ -26,11 +26,25 @@ jobs:
...
+        uses: actions/cache@v2
+        with:
+          path: ~/.ccache
+          # We include the commit sha in the cache key, as new cache entries are
+          # only created if there is no existing entry for the key yet.
+          key: ${{ runner.os }}-ccache-${{ github.sha }}
+          # Restore any ccache cache entry, if none for
+          # ${{ runner.os }}-ccache-${{ github.sha }} exists
+          restore-keys: |
+            ${{ runner.os }}-ccache
```
szha, line 39: if the cache key includes commit sha, who will create the cache without the commit cache key? how does this work?

sxjscience: @leezu I copied the code from the MXNet github actions. Am I using it correctly?

.github/workflows/unittests.yml
```diff
@@ -44,6 +58,21 @@ jobs:
...
+      - name: Build and Install TVM
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          git clone https://github.com/apache/incubator-tvm tvm --recursive
+          cd tvm
+          mkdir -p build
+          cp cmake/config.cmake build
+          echo set\(USE_LLVM ON\) >> build/config.cmake
+          echo set\(USE_GRAPH_RUNTIME ON\) >> build/config.cmake
+          echo set\(USE_BLAS openblas\) >> build/config.cmake
```
szha, line 70: do we need to turn on use cuda?

sxjscience:
It’s on CPU so no.

.github/workflows/unittests.yml
leezu: https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows#matching-a-cache-key

> You can provide a list of restore keys to use when there is a cache miss on key. You can create multiple restore keys ordered from the most specific to least specific. The cache action searches for restore-keys in sequential order. **When a key doesn't match directly, the action searches for keys prefixed with the restore key. If there are multiple partial matches for a restore key, the action returns the most recently created cache.**

.github/workflows/unittests.yml
leezu: I mean that you may not need to add the llvm apt repo: `sudo bash -c "$(wget -O - https://apt.llvm.org/llvm.sh)"`

leezu: > LLVM 4.0 or higher is needed for build with LLVM. Note that version of LLVM from default apt may lower than 4.0.

You don't need to add the upstream llvm repo. You can just use the llvm version available on Ubuntu itself. It may faster and less flaky (upstream website may be more likely to be unavailable at times).

.github/workflows/unittests.yml
sxjscience: It's on CPU so there is no need to do this.

sxjscience:
@leezu I’ve moved the LLVM installation code, and the CI has passed.

szha (approved):
thanks! this is an important integration

tqchen:
https://github.com/tlc-pack/tlcpack now provides nightly thirdparty build for tvm, packages can be installed from https://anaconda.org/tlcpack if you are using conda, this might be easier than building tvm from scratch every time

PR #1391, opened 2020-10-19
[BUGFIX] Fix Codecov

barry-jin:
## Description ##
Update coveragerc to omit scripts and tests when generating code coverage report on AWS Batch

### Essentials ###
- [ ] Code is well-documented

### Changes ###

## Comments ##

cc @dmlc/gluon-nlp-team


sxjscience (dismissed):

.coveragerc
```diff
@@ -9,6 +9,10 @@ concurrency =
 
 [report]
 ignore_errors = True
+omit = 
+    tests/*
+    scripts/*
```
leezu, line 14: Should we later include scripts again as part of the coverage?

sxjscience: I'm not sure. What's your opinion? Should we just include those that will be run as our scheduled test?

sxjscience (approved):
It seems that the problem has been fixed.

.coveragerc
leezu: Perhaps we can have a separate coverage report for the scripts. This will simplify the report generation as scripts testing schedule would differ from the general tests.

barry-jin:
> It seems that the problem has been fixed.

Not yet. GPU tests still cannot omit the coverage of tests/ and scripts/

PR #1399, opened 2020-10-25
[Fix] Improve test + Move TVM integration to experimental

sxjscience:
## Description ##
Fix seed for the TVM test, which can be flaky

cc @dmlc/gluon-nlp-team

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ process the text data, and train models.
 - Easy-to-use Text Processing Tools and APIs
 - Pretrained Model Zoo
 - Write Models with Numpy-like API
-- Fast Inference via [TVM](https://tvm.apache.org/) (Experimental)
+- Fast Inference via [Apache TVM (incubating)](https://tvm.apache.org/) (Experimental)
 - AWS Integration via [SageMaker](https://aws.amazon.com/sagemaker/)
 
 
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ process the text data, and train models.
 - Easy-to-use Text Processing Tools and APIs
 - Pretrained Model Zoo
 - Write Models with Numpy-like API
-- Fast Inference via [TVM](https://tvm.apache.org/)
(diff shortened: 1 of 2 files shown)


szha:
Fixing seed is usually not a good strategy as changes in upstream can still change the order of events and thus the results. If necessary, change the tolerance instead.

sxjscience:
Actually, the tolerance has been changed to 1E-1 and it can still be flaky. That’s the reason why I fixed the seed. We will need to investigate why it happens in TVM. One possibility is due to some numerical issues caused by the fusion.

szha:
This sounds like a real issue. Let’s proceed as you suggested and mark it experimental and note down the known issue

sxjscience:
I think TVM can still be used and may not be marked as experimental. Basically, the issue happens for the ALBERT network and is quite flaky. We need to spend some time to investigate the issue. I’ve also informed @icemelon9 about that.

szha:
We need to take the following into account:
- the TVM dependency is on an unreleased commit.
- (assuming TVM adopts semantic versioning) TVM hasn’t released a 1.0 yet, which means it doesn’t commit to a stable API.
- TVM integration is a new feature that was merged just now.
- GluonNLP supports production and needs to provide stability guarantee.

Please reconsider

sxjscience:
Marking it as experimental is also valid (The MXNet 2.0 might be itself experimental...... ). Nevertheless, we may encourage our users to try out TVM. According to the current benchmark, TVM can offer 1.2x -- 2.4x speed-up over huggingface + torchscript for fp32 inference. Thus, it is a nice feature to add in our package. We can later provide an example about how to convert GluonNLP models to TVM and performance inference on different AWS instances, including C4, C5, G4, P3. In addition, the inference latency can largely benefit from new features like Ansor.

sxjscience:
@szha Changed the README to point out that TVM integration is experimental. We may change it back once TVM 0.8 is ready.

szha (approved):
Approved as temporary fix

PR #1405, opened 2020-10-26
Add AMP + Update Benchmarking Script

sxjscience:
## Description ##
Add AMP support + update benchmarking script.

- [x] Testing utility of fp16 backbone
- [x] Update tests and benchmarking script
- [x] Update SQuAD finetuning to use amp
- [x] Use boolean masking

Some issues:

- The test for GPT-2 is turned off due to https://github.com/apache/incubator-mxnet/issues/19463
- The test of MobileBERT is turned off since it has numerical isssues in float16.
- TVM test has been disabled because it has some problems when we turned on boolean masking: https://github.com/dmlc/gluon-nlp/issues/1425 . We need to fix it on the TVM side.

### Essentials ###
- [x] Code is well-documented

cc @dmlc/gluon-nlp-team


scripts/question_answering/run_squad.py
```diff
@@ -527,7 +527,7 @@ def train(args):
                         'wd': args.wd,
                         'lr_scheduler': lr_scheduler,
                         }
-    adam_betas = eval(args.adam_betas)
+    adam_betas = json.loads(args.adam_betas.replace('(', '[').replace(')', ']'))
```
ZheyuYe, line 530: This seems a lot more complicated

sxjscience: Revised to use `ast.literal_eval`.

barry-jin (approved):
LGTM

scripts/machine_translation/README.md
```diff
@@ -30,6 +30,7 @@ python3 train_transformer.py \
     --save_dir transformer_base_wmt2014_en_de_${SUBWORD_ALGO} \
     --cfg transformer_base \
     --lr 0.002 \
+    --num_accumulated 32 \
```
szha, line 33: hmm was this missed from before?

sxjscience: It's the default argument. Just add it here to make it more explicit.

scripts/question_answering/README.md
```diff
@@ -184,3 +184,49 @@ For reference, we have also included the results of original version from Google
 |Google ELECTRA large      |      -/89.7     |     -/88.1   |
 |Google MobileBERT         |   90.0/82.9	|   79.2/76.2   |
 |Fairseq RoBERTa large     |   94.6/88.9    |	89.4/86.5   |
+
+### Run with AWS Batch
+We can quickly run the squad finetuning via the [AWS Batch support](../../tools/batch).
```
szha, line 189: is the batch support refactored so that it can be bootstrapped in *any* AWS account? including VPC setup, security group, etc?

szha: if it doesn't work we should avoid advertising this as user facing feature.

scripts/question_answering/batch/sync_batch_result.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -ex
+
+LOG_PATH=$1
+SAVE_DIR_NAME=${2:-squad_2.0}
+
+while read -r job_name job_id; do
+    aws s3 sync s3://gluon-nlp-dev/batch/${job_id}/temp ${SAVE_DIR_NAME}/${job_name}
+done < ${LOG_PATH}
```
szha, line 10: I don't think we should check in all the convenience scripts in the scripts, because this folder is user-facing.

scripts/question_answering/README.md
sxjscience: I think it's for the purpose of running the experiments quickly if you have batch access. Otherwise, we will end up writing everything inside the batch folder.

scripts/question_answering/batch/sync_batch_result.sh
sxjscience: For me, I feel that it's fine to help reproduce the experiments for our own purpose. And also later add to the scheduled check.

scripts/question_answering/README.md
szha: it's also ok to just keep the script somewhere local or in a gist.

szha (dismissed):
let's remove the non-user-facing scripts from the user facing scripts folder.

scripts/question_answering/README.md
sxjscience: I've also considered this. The problem is that it will increase the communication time and it can add some values to the user if we will later have the cloudformation script.

szha: if the feature doesn't work for a user as is, it will cause confusion and friction.

szha: keeping it as part of infra code is also ok.

sxjscience: May be let's still keep it in the batch folder.

scripts/question_answering/batch/sync_batch_result.sh
sxjscience: Should be resolved now.

leezu (approved):

PR #1427, opened 2020-11-05
Remove gluonnlp.layers.HybridSequential

leezu:
As title. No longer needed as of https://github.com/apache/incubator-mxnet/commit/3d1df4e035077b5d4de859e7bf464181388d47a0


sxjscience (approved):
LGTM. We may fix the doc error later.

szha (approved):

PR #1448, opened 2020-12-02
[AMP] Add AMP support to Machine Translation

sxjscience:
## Description ##
- Fix the horovod support and add the amp support to machine translation. 
- Support TN in training and inference
- Update training results of SQuAD, transformer-base, transformer-large, transformer-t2t-big.
- Add Deep Encoder, Shallow Decoder

### Essentials ###
- [x] Code is well-documented

### Changes ###
- [x] Add AMP, tests

## Comments ##

cc @dmlc/gluon-nlp-team


sxjscience:
@leezu @szhengac @xinyual 

Also thanks Xinyuan in trying out the Hyper-params. I think we can continue improving this version.

scripts/machine_translation/train_transformer.py
```diff
@@ -422,36 +425,50 @@ def train(args):
...
-        while not is_last_batch:
-            processed_batch_num += len(sample_data_l)
+    log_avg_loss_l = [mx.np.array(0.0, ctx=ctx) for ctx in ctx_l]
+    log_avg_loss_counter = 0
+    log_wc_l = [mx.np.array(0, dtype=np.int64, ctx=ctx) for ctx in ctx_l]
+
+    if local_rank == 0:
+        writer = SummaryWriter(logdir=os.path.join(args.save_dir, 'tensorboard'))
+    if use_amp:
+        amp.init_trainer(trainer)
```
szhengac, line 435: We should allow the users to set the initial scaling, e.g., 2^20.

sxjscience: I checked the API and find that there is no such option in the amp in MXNet. Is it necessary for large batch?

scripts/machine_translation/README.md
```diff
@@ -160,11 +190,56 @@ Test BLEU score with 3 seeds (evaluated via sacre BLEU):
...
 |---------------|-------------|
-| yttm          | 26.78/25.96 |
+| yttm          | 26.95/25.85 |
 
 - transformer_wmt_en_de_big
 
 (test bleu / valid bleu)
 | Subword Model | Seed = 123  |
 |---------------|-------------|
 | yttm          | 27.99/26.84 |
```
xinyual, line 200: I try big NMT. The lr 0.001 will diverge so I decrease it to 0.0005 and the result on testset is 27.818.

scripts/machine_translation/train_transformer.py
szhengac: This is something necessary.

szhengac: And, the ability of tuning scaling factor, window size, etc is also needed.

sxjscience: I see. What would be the typical flag in pytorch AMP training? I think the MXNet AMP has not offered the functionality of setting the options. Need to confirm with NVIDIA.

src/gluonnlp/utils/parameter.py
```diff
@@ -180,3 +181,62 @@ def move_to_ctx(arr, ctx):
         return [move_to_ctx(ele, ctx) for ele in arr]
     else:
         return None if arr is None else arr.as_in_ctx(ctx)
+
+
+def deduplicate_param_dict(param_dict):
```
sxjscience, line 186: @leezu @xinyual @szha  I added this `deduplicate_param_dict` because I find that `model.collect_params()` will return all parameters, even if some are shared. Thus, if there are shared parameters, `allreduce` will be called for multiple times. The results will thus be incorrect.

sxjscience:
@szhengac @leezu @szha Should be ready for merge now. I'm still running the Deep Encoder + Shallow Decoder model and decide to attach the result in a later PR. The AMP training results

- transformer-base: 27.07 BLEU
- transformer-large: 28.00 BLEU
- transformer-big-t2t: 28.00 BLEU

Also cc @MoisesHer .

sxjscience:
Updated the `enc12-dec1` results. It achieved 26.2 sacreBLEU score.

scripts/generation/README.md
```diff
@@ -1,5 +1,7 @@
-#Generation
+#Sequence Generation
```
szha, line 1: nit: put a space after `#`

scripts/machine_translation/README.md
```diff
@@ -1,170 +1,328 @@
 # Machine Translation
 
+To run the scripts, you are recommended to install [tensorboardX](https://github.com/lanpa/tensorboardX).
+
+```
+python3 -m pip install tensorboardX
```
szha, line 6: out of curiosity, is tensorboardx still necessary for using tensorboard without tensorflow?

scripts/machine_translation/README.md
```diff
@@ -1,170 +1,328 @@
...
 
 ```bash
 SUBWORD_ALGO=yttm
 SRC=en
 TGT=de
-python3 train_transformer.py \
+lr=0.0016
+num_accumulated=16
+max_num_tokens=4096
+epochs=60
```
szha, line 38: all uppercase to be consistent with the other parameterized arguments?

scripts/machine_translation/evaluate_epochs_wmt2014_ende.sh
```diff
@@ -0,0 +1,50 @@
...
+        --tgt_lang de \
+        --cfg ${SAVE_DIR}/config.yml \
+        --src_tokenizer ${SUBWORD_ALGO} \
+        --tgt_tokenizer ${SUBWORD_ALGO} \
+        --src_subword_model_path wmt2014_ende/${SUBWORD_ALGO}.model \
+        --tgt_subword_model_path wmt2014_ende/${SUBWORD_ALGO}.model \
+        --src_vocab_path wmt2014_ende/${SUBWORD_ALGO}.vocab \
+        --tgt_vocab_path wmt2014_ende/${SUBWORD_ALGO}.vocab \
+        --src_corpus wmt2014_ende/test.raw.en \
+        --tgt_corpus wmt2014_ende/test.raw.de \
```
szha, line 25: nit: would it make sense to parametrize dev/test instead of duplicate the commands? either is ok

scripts/machine_translation/evaluate_transformer.py
```diff
@@ -29,14 +30,29 @@ def parse_args():
                         help='The source corpus for evaluation.')
     parser.add_argument('--tgt_corpus', type=str, default=None,
                         help='The target corpus for evaluation.')
+    parser.add_argument('--src_normalizer', choices=['no', 'moses'],
+                        default='moses', help='The sentence normalizer that will be '
+                                              'used to normalize the source sentence.')
```
szha, line 35: I'm not sure whether "normalize" is a helpful concept, as it doesn't tell me what it does

szha: if it's already a convention elsewhere I'm ok to learn that (so that it's easier to communicate with that group). Just want to point out the semantics of "normalize" isn't as clear as "tokenize"

scripts/machine_translation/README.md
sxjscience: Not sure. I haven't checked the latest version of Tensorboard. But in the official documentation of Tensorboard, there is only the use-case within TF.

scripts/machine_translation/README.md
sxjscience: Both are fine for me.

scripts/machine_translation/evaluate_transformer.py
sxjscience: Usually, it means some pre-filtering steps before tokenization. For example, the definition of lossless tokenization is 

```
Decode(Encode(normalize(text))) == normalize(text)
```

Ref: Sentencepiece paper: https://arxiv.org/pdf/1808.06226.pdf, Page2.

scripts/machine_translation/train_transformer.py
```diff
@@ -348,35 +487,56 @@ def train(args):
...
                                                     from_logits=False)
     label_smooth_loss.hybridize()
-    rescale_loss = 100.0
-    
+
+    # Construct the beam search sampler
+    scorer = BeamSearchScorer(alpha=args.lp_alpha,
+                              K=args.lp_k,
+                              from_logits=False)
+    beam_search_sampler = BeamSearchSampler(beam_size=args.beam_size,
```
szha, line 495: I wonder if this should be done outside of `def train`. Current training is done through MLE so decoding doesn't seem needed.

sxjscience: It's currently included because we will need to report the validation BLEU score.

scripts/machine_translation/evaluate_epochs_wmt2014_ende.sh
sxjscience: Fixed.

src/gluonnlp/utils/parameter.py
```diff
@@ -180,3 +277,66 @@ def move_to_ctx(arr, ctx):
...
+    param_dict
+        The parameter dict returned by `model.collect_params()`
+
+    Returns
+    -------
+    dedup_param_dict
+    """
+    dedup_param_dict = dict()
+    param_uuid_set = set()
+    for k, v in param_dict.items():
```
szha, line 296: the key that will be kept in the resulting dictionary will not be deterministic because of the potentially different iteration order. I'd suggest only returning the value so that we won't run into inconsistent dictionary problem across workers due to incorrect assumption.

scripts/generation/README.md
sxjscience: Fixed.

src/gluonnlp/utils/parameter.py
szha: an alternative would be to iterate with sorted keys.

scripts/question_answering/run_squad.py
```diff
@@ -491,7 +493,7 @@ def train(args):
 
     logging.info('Creating distributed trainer...')
     # Collect differentiable parameters
-    param_dict = qa_net.collect_params()
+    param_dict = deduplicate_param_dict(qa_net.collect_params())
```
szha, line 496: does the result make more sense after correcting for duplication?

src/gluonnlp/utils/parameter.py
sxjscience: OK, I thought that starting from python3.6, this will be deterministic.

szha (commented):
Half-way through review now. So far so good. I will resume a bit later.

scripts/question_answering/run_squad.py
sxjscience: The results are better I think. The main reason is that it fixed the gradient norm calculation. Also, it does not have an impact on the MXNet part because it has been deduplicated already in MXNet (as confirmed with @leezu ).

src/gluonnlp/utils/parameter.py
sxjscience: Changed to use `sorted`.

scripts/machine_translation/train_transformer.py
szha: should it be part of evaluation method?

scripts/datasets/general_nlp_benchmark/prepare_glue.py
```diff
@@ -467,20 +467,20 @@ def read_broadcoverage_diagnostic(dir_path):
 
 
 GLUE_TASK2PATH = {
-    "cola": "https://firebasestorage.googleapis.com/v0/b/mtl-sentence-representations.appspot.com/o/data%2FCoLA.zip?alt=media&token=46d5e637-3411-4188-bc44-5809b5bfb5f4",  # noqa
-    "sst": "https://firebasestorage.googleapis.com/v0/b/mtl-sentence-representations.appspot.com/o/data%2FSST-2.zip?alt=media&token=aabc5f6b-e466-44a2-b9b4-cf6337f84ac8",  # noqa
+    "cola": "https://gluonnlp-numpy-data.s3-accelerate.amazonaws.com/datasets/text_classification/glue_superglue/glue/cola.zip",  # noqa
+    "sst": "https://gluonnlp-numpy-data.s3-accelerate.amazonaws.com/datasets/text_classification/glue_superglue/glue/sst.zip",  # noqa
```
szha, line 471: for our own mirrors, I recommend to always follow the naming convention where a content hash prefix is added as a suffix to the base name. The reason is that this will ensure updating data files will not affect results of older scripts.

szha: in the current setting where only hash verification happens, we won't be able to update the files without making files that contain old content hashes unrunnable.

tools/batch/question_answering/parse_squad_results.py
```diff
@@ -14,7 +14,7 @@
 args = parser.parse_args()
 
 if args.save_path is None:
-    args.save_path = os.path.basename(os.path.dirname(args.dir)) + '.csv'
```
szha, line 17: if `--dir` is pointing to a directory, should the basename of the saved file come from somewhere else?

scripts/datasets/general_nlp_benchmark/prepare_glue.py
sxjscience: I think these are fixed and will not change (Because these are GLUE + SuperGLUE datasets). It should be fine to not add the hash prefix for these files.

tools/batch/question_answering/parse_squad_results.py
sxjscience: The idea is actually to use `DIR_NAME.csv` as the file name.

scripts/machine_translation/train_transformer.py
sxjscience: Since validation is part of the training process, it should be better to be created in the train function.

scripts/machine_translation/README.md
szhengac: tensorboard should be ok now.

szhengac (approved):

src/gluonnlp/sequence_sampler.py
```diff
@@ -286,22 +285,26 @@ def gumbel_with_maximum(self, phi, T, dim=-1):
...
         ----------
-        F
         g_phi : mx.np.ndarray
             Shape (batch_size, beam_size, L).
         T : mx.np.ndarray
             The previous scores. Shape (batch_size, beam_size)
+        axis
+            The axis
+        Z
+            The Z value
```
szha, line 300: we can probably make these descriptions more helpful. the docstring above may also help.

src/gluonnlp/utils/misc.py
```diff
@@ -332,6 +225,7 @@ def repeat(iterable, count=None):
             for sample in iterable:
                 yield sample
 
+
 def parse_ctx(data_str):
```
szha, line 229: make this private or add a docstring

src/gluonnlp/utils/misc.py
```diff
@@ -610,12 +504,16 @@ def init_comm(backend, gpus):
     Parameters
     ----------
     backend
+        The communication backend
     gpus
```
szha, line 508: docstring

tests/test_data_sampler.py
```diff
@@ -136,12 +136,14 @@ def test_split_sampler_even_size(num_samples, num_parts):
 @pytest.mark.parametrize('max_num_tokens', [200, 500])
 @pytest.mark.parametrize('max_num_sentences', [-1, 5])
 @pytest.mark.parametrize('required_batch_size_multiple', [1, 5])
+@pytest.mark.parametrize('sort_type', ['sequential', 'max'])
```
szha, line 139: first?

src/gluonnlp/utils/parameter.py
szha: @leezu shall we consider allowing deduplication in block.collect_params? we currently have regex as a filter already and enhancing filter logic with deduplication seems to make sense.

szha (approved):
Overall looks good to me. Most of the remaining comments are on missing docstrings, which I understand is already being tracked so I'm ok with having those changes as follow-ups.

PR #1459, opened 2020-12-17
Unblock CI by skipping the test

sxjscience:
## Description ##
(Brief description on what this PR is about)

### Essentials ###
- [ ] Code is well-documented

### Changes ###

## Comments ##

cc @dmlc/gluon-nlp-team

### DIFF
--- a/tests/test_gluon_block.py
+++ b/tests/test_gluon_block.py
@@ -51,7 +51,8 @@ def test_gluon_nonzero_hybridize():
     out.wait_to_read()
 
 
-@pytest.mark.skip('Skip due to MXNet bug https://github.com/apache/incubator-mxnet/issues/19659')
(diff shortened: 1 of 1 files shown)


tests/test_gluon_block.py
```diff
@@ -50,6 +51,7 @@ def forward(self, x):
     out.wait_to_read()
 
 
+@pytest.mark.skip('Skip due to MXNet bug https://github.com/apache/incubator-mxnet/issues/19659')
```
szha, line 54: is it still worth running? we can also mark it as xfail.

sxjscience: It's worth running and may appear if we try to hybridize network with boolean mask.

szha: In that case I would recommend changing it to xfail mark, so that the test would be run (and expected to fail for now)

szhengac (approved):

PR #1467, opened 2021-01-07
Speed up downloads from S3 bucket

leezu:
Rely on boto3 by default for downloads from S3. Handle unauthenticated access to S3 via botocore if needed.

### DIFF
--- a/src/gluonnlp/utils/misc.py
+++ b/src/gluonnlp/utils/misc.py
@@ -363,8 +363,11 @@ def download(url: str,
     """
     is_s3 = url.startswith(S3_PREFIX)
     if is_s3:
-        boto3 = try_import_boto3()
+        boto3, botocore = try_import_boto3()
         s3 = boto3.resource('s3')
+        if boto3.session.Session().get_credentials() is None:
+            from botocore.handlers import disable_signing
+            s3.meta.client.meta.events.register('choose-signer.s3.*', disable_signing)
         components = url[len(S3_PREFIX):].split('/')
         if len(components) < 2:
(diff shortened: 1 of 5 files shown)


sxjscience (approved):
LGTM

sxjscience:
@Cli212 After this gets merged, we can directly add Natural Question to the GluonNLP Dataset.

src/gluonnlp/utils/misc.py
```diff
@@ -363,8 +363,11 @@ def download(url: str,
     """
     is_s3 = url.startswith(S3_PREFIX)
     if is_s3:
-        boto3 = try_import_boto3()
+        boto3, botocore = try_import_boto3()
         s3 = boto3.resource('s3')
+        if boto3.session.Session().get_credentials() is None:
+            from botocore.handlers import disable_signing
+            s3.meta.client.meta.events.register('choose-signer.s3.*', disable_signing)
```
szha, line 370: does this behavior need be parametrized in the download function?

leezu: What's your suggestion? The `s3` resource is created only in the download function

szha: I was thinking about having a `use_unsigned` argument so that anonymous downloading is explicit. Otherwise user may not be aware if the credential passing failed unexpectedly.

leezu: I think we're just working around a bug in boto, thus there is no value in exposing it to users. There is no reason for boto to fail to download public files if no credentials are passed.

PR #1470, opened 2021-01-08
adjust implementation of relative position encoding

yongyi-wu:
## Description ##

This version aligns the implementation of relative position encoding with that in mesh_tensorflow. It would also help program T5 model in the future. 

## Changes ##
- Define relative position as `mem_i - query_j` in lieu of previous `query_i - mem_j `
- Rephrase documentation

cc @dmlc/gluon-nlp-team

### DIFF
--- a/src/gluonnlp/op.py
+++ b/src/gluonnlp/op.py
@@ -231,16 +231,14 @@ def relative_position_bucket(relative_position,
                              bidirectional: bool = True,
                              num_buckets: int = 32,
                              max_distance: int = 128):
-    """Map the relative position to buckets. The major difference between our implementation and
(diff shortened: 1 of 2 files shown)


src/gluonnlp/op.py
```diff
@@ -249,8 +247,8 @@ def relative_position_bucket(relative_position,
         Shape (...,)
```
sxjscience, line 247: Would you also help to remove `F` here?

sxjscience (approved):
LGTM in general. If possible, you can also remove the `F`.

sxjscience:
We can try to improve the docstring later.

PR #1476, opened 2021-01-10
Fix typo in test_bert.py

leezu:
## Description ##
In `test_bert.py` we assert about `mlm_score` instead of `mlm_scores` in the latter part of the test.
To avoid similar copy-paste errors in the future, consistently use `mlm_score` in the tests.


sxjscience (approved):
LGTM

PR #1482, opened 2021-01-15
[FEATURE] Add T5Model and T5 conversion scripts; Adjust T5Tokenizer instantiation

yongyi-wu:
## Description ##
This PR contributes the T5 model to the GluonNLP codebase with scripts that could convert a Huggingface T5 model. It also extends functionalities of some modules in layer.py. Finally, this PR enables user to create a T5 tokenizer with control of `<extra_id>` tokens' quantity. 

### Essentials ###

### Changes ###
- [x] `T5Model` inplementation (forward pass tested)
- [x] Conversion scripts for Huggingface `T5Model` with conversion results uploaded to S3
- [x] Improved `SentencepieceTokenizer` instantiation
- [x] `RMSNorm` and Gated-`PositionwiseFFN` in layer.py
- [x] Bug and typo fixes

cc @dmlc/gluon-nlp-team


scripts/conversion_toolkits/convert_t5.py
```diff
@@ -0,0 +1,100 @@
+
+
+def parse_args(): 
+    pass
```
sxjscience, line 4: Let's try to add the argparse and add the huggingface model name as an argument, e.g., t5-small, t5-base, t5-large.

src/gluonnlp/models/t5.py
```diff
@@ -0,0 +1,849 @@
...
+    cfg.MODEL.d_model = 1024
+    cfg.MODEL.d_kv = 128
+    cfg.MODEL.d_ff = 65536
+    cfg.MODEL.num_layers = 24
+    cfg.MODEL.num_heads = 128
+    cfg.freeze()
+    return cfg
+
+
+PRETRAINED_URL = {
```
sxjscience, line 138: Let's upload the models to 
```
s3://gluonnlp-numpy-data/models/
```

sxjscience:
@ZheyuYe Would you help review it also if you have time? Also ping @szhengac to help review if you have time.

sxjscience:
I told @yongyi-wu to implement incremental_decode in a follow-up PR and we will focus on the correctness of the forward pass first in this PR.

yongyi-wu:
I am working on convert_t5.py and uploading model weights to S3, as @sxjscience noted previously. I will update here when I'm done.

src/gluonnlp/models/t5.py
```diff
@@ -0,0 +1,849 @@
...
+            return 0, 0
+        else: 
+            return 1, 1
+
+    @_assert_decoder_method
+    def _init_states(self, batch_size, ctx, dtype='float32'): 
+        if self.layout == 'NT': 
+            shape = (batch_size, 0, self._num_heads, self._d_kv)
+        else: 
+            shape = ((0, batch_size, self._num_heads, self._d_kv))
```
sxjscience, line 438: Need to drop the additional tuple:

`((0, batch_size, self._num_heads, self._d_kv))` --> `(0, batch_size, self._num_heads, self._d_kv)`

src/gluonnlp/models/t5.py
```diff
@@ -0,0 +1,849 @@
...
+        past_key_value, 
+        mem_states, 
+        mem_valid_length
+    ): 
+        raise NotImplementedError
+
+    def _get_query_key_shape(self, hidden_states, mem_states, past_key_value): 
+        # NT: (B, L_seq, inner_dim); TN: (L_seq, B, inner_dim)
+        index = 1 if self.layout == 'NT' else 0
+        query_length = hidden_states.shape[index]
```
sxjscience, line 604: This will make it not be able to hybridize.

src/gluonnlp/models/t5.py
```diff
@@ -0,0 +1,849 @@
...
+            # for incremental decoding only, where past key and value are of shape
+            # NT(NTK): (B, L_seq, num_heads, n_kv); TN(TNK): (L_seq, B, num_heads, n_kv)
+            query_length += past_key_value[0].shape[index]
+        key_length = query_length if mem_states is None else mem_states.shape[index]
+        return query_length, key_length
+
+    @staticmethod
+    def _get_relative_position(query_length, key_length): 
+        # relative_position = mem_i - query_j
+        query_position = np.arange(query_length, dtype=np.int32)[:, None]
```
sxjscience, line 615: For example, assume data has shape `(B, query_length, mem_length)`, then `npx.arange_like(hidden_states, axis=1)` will be `[0, 1, 2, ..., query_length - 1]`.

src/gluonnlp/models/t5.py
```diff
@@ -0,0 +1,849 @@
...
+    ): 
+        # 1. relative position embeddings and attention masks
+        position_embeddings = self.relative_position_encoder(
+            T5Stack._get_relative_position(*self._get_query_key_shape(hidden_states, None, None))
+        )
+        if self._is_decoder: 
+            mem_position_embeddings = np.zeros(
+                (1, self._num_heads, *self._get_query_key_shape(hidden_states, mem_states, None)), 
+                dtype=self._dtype
+            )
```
sxjscience, line 635: Try to use zeros_like.

sxjscience: You may try to first use `arange_like` to get the two ndarrays, one with shape (query_length,), and another with shape (mem_length,). After that, you can use these two arrays to get `(query_length, mem_length)` and later get `(1, num_heads, query_length, mem_length)`.

src/gluonnlp/models/t5.py
```diff
@@ -0,0 +1,849 @@
...
+        dropout_prob=0.1, 
+        layer_norm_eps=1E-6, 
+        activation='relu', 
+        init_factor=1.0, 
+        dtype='float32'
+    ): 
+        super().__init__()
+        self._activation = activation
+
+        if activation == 'relu':
```
ZheyuYe, line 196: for this the conditional statements `if ... elif .. else` here, introducing an additional projection layer `wi_1` is sufficient

src/gluonnlp/models/t5.py
```diff
@@ -0,0 +1,849 @@
...
+        """
+        hidden_states: 
+            - layout = 'NT'
+                Shape (B, L_seq, d_model)
+            - layout = 'TN'
+                Shape (L_seq, B, d_model)
+        """
+        # NT -> NTK: (B, L_seq, inner_dim) -> (B, L_seq, num_heads, n_kv)
+        # TN -> TNK: (L_seq, B, inner_dim) -> (L_seq, B, num_heads, n_kv)
+        def shape(x):
```
ZheyuYe, line 472: Could you please rename this function to avoid ambiguity

yongyi-wu: I kept this consistent with what huggingface has in its implementation, as I couldn't come up with a better name. Do you have any suggestion? 
FYI: https://github.com/huggingface/transformers/blob/c60e0e1ee45f4bf1017736b146c51729f120bb83/src/transformers/models/t5/modeling_t5.py#L437

ZheyuYe: Can we refer to the name Google used in the bert source code as https://github.com/google-research/bert/blob/eedf5716ce1268e56f0a50264a88cafad334ac61/modeling.py#L629

yongyi-wu: Sure, that sounds awesome.

src/gluonnlp/models/t5.py
yongyi-wu: I used `expand_dims` instead and now the hybridization works fine. Also, `hybridize` seems to support `.shape` now. Thank you for pointing it out though.

src/gluonnlp/models/t5.py
szha: It should be able to under deferred compute, provided that it doesn't change in the hybridized model. cc @leezu to confirm

sxjscience: The length of the key and query may change and I think it cannot deal with that even with deferred compute?

sxjscience: @szha If that’s the case, it should be advertised. For me, I feel that the final symbol will be the same even if we have changed query_length and key_length.

szha: In deferred compute, the shape information is accessible as shape inference happens eagerly. If that value gets passed into a node attribute, then that attribute is frozen in the hybridized graph.

For example, suppose the query length is then used in normalization `x/length`, the operator invoked would be a broadcast scalar division, where the node attribute value of the divisor is from length, a python integer from `.shape[index]`. In that case, as the length changes, the division value would be wrong. To address it, one can use `.shape_array[index]` so that the op invoked is broadcast division and it's always the up-to-date length value that will be used.

sxjscience: The current problem is that we will need to use query_length, key_length to create new matrices. For example, mx.np.arange(query_length).

sxjscience: Is this also supported? I think it should not be runnable.

szha: Not yet. It will be possible once we get to creation ops with symbolic shapes. (cc @barry-jin about the example use case above)

leezu: > I think it should not be runnable.

A error will be raised once you input shapes that would yield a different intermediate matrix

sxjscience: > > I think it should not be runnable.
> 
> A error will be raised once you input shapes that would yield a incompatible intermediate matrix

Yes, this simply indicates that this feature is not supported yet and we should not recommend users to use `.shape` in forward.

src/gluonnlp/models/t5.py
yongyi-wu: That's a great point. I am now integrating this part into the `PositionwiseFFN` of layer.py.

src/gluonnlp/models/t5.py
```diff
@@ -0,0 +1,849 @@
...
+
+    }
+}
+
+
+# FILE_STATS = load_checksum_stats(os.path.join(get_model_zoo_checksum_dir(), 't5.txt'))
+
+
+@use_np
+class T5LayerNorm(HybridBlock):
```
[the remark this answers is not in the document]
yongyi-wu: T5's layer normalization is a bit weird in that it does not do mean substraction nor use bias, so I'm not sure whether it is likely to be re-used later. I can absolutely move it to `..layer` though if you know this module can be used elsewhere.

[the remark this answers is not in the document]
sxjscience: I think they are using RMSNorm: https://arxiv.org/pdf/1910.07467.pdf We can add that to layers.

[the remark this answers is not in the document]
yongyi-wu: Cool, I'll go with that.

yongyi-wu:
Hi all, thank you for your patience. I've updated the code based on your feedback. At this point, `google_t5_small` is up on S3 already, whereas other files should be uploaded shortly. By the way, there are some non-trivial changes in `t5.py` as compared to the previous commits, so it would help a lot if you could review it again :D

scripts/conversion_toolkits/convert_t5.sh
```diff
@@ -0,0 +1,7 @@
+python3 -m pip install git+https://github.com/huggingface/transformers.git --upgrade
+for model in small base large 3B 11B
+do
+    dest_dir="google_t5_${model}"
+    mkdir ${dest_dir}
+    python3 convert_t5.py "t5-${model,,}" ${dest_dir} --test
+done
```
sxjscience, line 7: nit: new line.

src/gluonnlp/models/t5.py
```diff
@@ -0,0 +1,883 @@
...
+            dtype=dtype
+        )
+        # avoid scaling before softmax
+        # See https://github.com/tensorflow/mesh/blob/fa19d69eafc9a482aff0b59ddd96b025c0cb207d/mesh_tensorflow/transformer/attention.py#L136
+        self.self_attn_q = nn.Dense(
+                units=self._inner_dim, 
+                in_units=d_model, 
+                flatten=False, 
+                use_bias=False, 
+                weight_initializer=Normal((d_model * d_kv) ** -0.5 * init_factor),
```
sxjscience, line 209: @ZheyuYe @leezu @szhengac @szha Just mention here in case you do not konw. T5 used this trick to avoid dividing the `sqrt(query_dim)` as in `softmax(<query, key> / sqrt(query_dim))`. Basically, the trick is to move `1/sqrt(query_dim)` to the initializer.

szha: They don't seem equivalent

yongyi-wu: I leave it consistent with huggingface's implementation.

sxjscience: Since all query are generated as `W_q X`. Calculating `W_q X / sqrt(query_dim)` is equivalent to `W_q / sqrt(query_dim) X`. We can then move `1/sqrt(query_dim)` to the initialization phase of `W_q` so that they will be the same when we first initialize the model.

sxjscience: @szha Would you give more detail about the difference between these two approaches? I feel that they give us the same initialization of the attention scores.

sxjscience (approved):
Looks good to me. A very nice addition! Let's try to convert 3B + 11B when we are fixing the CI. Basically, T5-3B + T5-11B are quite large and they can help test the large-tensor support in MXNet. If we met any issues, they might be related to bugs in MXNet.

src/gluonnlp/layers.py
```diff
@@ -144,6 +146,40 @@ def __repr__(self):
...
 
 
+@use_np
+class RMSNorm(HybridBlock): 
+    """Apply root mean square layer normalization to n-dimensional input array, 
+    where we do not substract mean in the numerator. 
+
+    For more details, see the paper: https://arxiv.org/pdf/1910.07467.pdf
+    """
+    def __init__(self, in_channels, center=True, scale=True,
```
sxjscience, line 156: We may change center to `use_bias`. and then set it to `False` by default to align with T5.

sxjscience:
Some other follow-up items:

- Convert the T5ConditionalGenerator so that we can directly use it to fill in missing values in the input sentence
- Add incremental_decode to support BeamSearch and fast inference.

scripts/conversion_toolkits/convert_t5.py
```diff
@@ -0,0 +1,266 @@
...
+    "t5-3b": "https://huggingface.co/t5-3b/resolve/main/config.json",
+    "t5-11b": "https://huggingface.co/t5-11b/resolve/main/config.json"
+}
+PRETRAINED_VOCAB_MAP = {
+    "t5-small": "https://huggingface.co/t5-small/resolve/main/spiece.model",
+    "t5-base": "https://huggingface.co/t5-base/resolve/main/spiece.model",
+    "t5-large": "https://huggingface.co/t5-large/resolve/main/spiece.model",
+    "t5-3b": "https://huggingface.co/t5-3b/resolve/main/spiece.model",
+    "t5-11b": "https://huggingface.co/t5-11b/resolve/main/spiece.model"
+}
```
sxjscience, line 39: We may consider to later rely only on `from_pretrained()` for the conversion.

src/gluonnlp/layers.py
sxjscience: Resolve it because I noticed that we have been using `center` in LayerNorm.

src/gluonnlp/models/t5.py
szha: The obvious difference is that the gradients would be of different scales.

sxjscience: Have you seen any difference other than the scale of the gradient? For me, I feel that the main point is that this ensures that the softmax attention weights are initialized similarly.

szha: I haven't tried this myself and the following is just based on reasoning. Since the gradients are different, I'd expect this parameter to be trained differently beyond initialization. In your example, the new gradient of W_q after change would be `X` instead of `X / sqrt(query_dim)`. Then, behaviors such as clipping global gradient norm and regularization will change too.

yongyi-wu:
> Looks good to me. A very nice addition! Let's try to convert 3B + 11B when we are fixing the CI. Basically, T5-3B + T5-11B are quite large and they can help test the large-tensor support in MXNet. If we met any issues, they might be related to bugs in MXNet.

T5-3B and T5-11B are good to go. Previous failure was probably due to OOM, while MXNet itself worked well.

sxjscience:
Let's also add the test here: https://github.com/dmlc/gluon-nlp/blob/4b061305285f3c4e491b77b046a92c32e306cbc6/tests/test_models.py#L26-L39 We may skip the t5-large, t5-3B, t5-11B and only test for small and base.

sxjscience:
@yongyi-wu After a second thought, I think we can skip t5 in test_models.py and just test it in the `test_models_t5.py`. We may try to load the `t5-small` and `t5-base` and do a basic forward. This helps us check if we can correctly load the model.

yongyi-wu:
> @yongyi-wu After a second thought, I think we can skip t5 in test_models.py and just test it in the `test_models_t5.py`. We may try to load the `t5-small` and `t5-base` and do a basic forward. This helps us check if we can correctly load the model.

After discussing with @sxjscience, we decided to still test T5 backbones in `test_models.py`, but skip T5-3B and T5-11B.

Issue #1481, opened 2021-01-15
[Bug] Bug of top-p sampling


sxjscience:
## Description
The top-p sampling here is not correct. https://github.com/dmlc/gluon-nlp/blob/52da8aba758b826189505213e1c01f12ebd944fd/src/gluonnlp/sequence_sampler.py#L720-L724 We should sort the probabilities in descending order and then do a cumsum of the probabilities. After that, we set the threshold based on the cumsum values.

@hymzoque

PR #1503, opened 2021-01-26
[BUGFIX]fix bug of top-p sampling

hymzoque:
## Description ##
fix bug of top-p sampling mentioned in [issue](https://github.com/dmlc/gluon-nlp/issues/1481)

### Essentials ###
- [ ] Code is well-documented

cc @dmlc/gluon-nlp-team

### DIFF
--- a/scripts/generation/README.md
+++ b/scripts/generation/README.md
@@ -14,6 +14,7 @@ python3 generate_unconditional_gpt2_samples.py \
     --gpu 0 \
     --temperature 0.7 \
     --top_k 40 \
+    --batch_size 2 \
     --nsamples 1000 > samples
 ```
 
@@ -43,7 +44,7 @@ Some metrics for the unconditional generated text
 | topk=40       | 0.4291         | 0.9666         | 0.0            |
 | topk=640      | 0.3384         | 0.9623         | 0.0            |
 | topk=40 t=0.7 | 0.4621         | 0.9586         | 1.1            |
-
+| topp=0.95     | 0.2676         | 0.9519         | 0.0            |
 
 Part of some interesting generated unconditional example
 
--- a/src/gluonnlp/sequence_sampler.py
+++ b/src/gluonnlp/sequence_sampler.py
@@ -727,7 +727,6 @@ class _MultinomialStepUpdate(HybridBlock):
              # choose the borderline prob
              p_prob = mx.np.min(masked_probs, axis=2, keepdims=True)
              probs = mx.np.where(
-                 probs > self._sampling_topp,
                  probs >= p_prob,
                  probs,
                  mx.np.zeros_like(probs)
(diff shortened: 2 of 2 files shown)


scripts/generation/README.md
```diff
@@ -43,7 +44,7 @@ Some metrics for the unconditional generated text
 | topk=40       | 0.4291         | 0.9666         | 0.0            |
```
sxjscience, line 44: I think previously we have the results of `t=0.9`, we should remove that row.

src/gluonnlp/sequence_sampler.py
```diff
@@ -717,11 +717,20 @@ def forward(self, samples, valid_length, outputs, scores, step, beam_alive_mask,
...
-            )
+             sorted_probs, sorted_indices = mx.npx.topk(probs, axis=2, k=-1, ret_typ='both', is_ascend=False)
+             cumsum_probs = mx.np.cumsum(sorted_probs, axis=2)
+             masked_probs = mx.np.where(
+                 cumsum_probs > self._sampling_topp,
+                 sorted_probs,
+                 mx.np.zeros_like(probs)
+             )
+             # choose the borderline prob
+             p_prob = mx.np.min(masked_probs, axis=2, keepdims=True)
```
sxjscience, line 728: Is it possible to use exactly the same implementation as https://gist.github.com/thomwolf/1a5a29f6962089e871b94cbd09daf317?

sxjscience:
@szha Would you help review?

sxjscience:
@szha Would you take a look?

src/gluonnlp/sequence_sampler.py
szha: I don't think sort is implemented yet.

sxjscience: I'm referring to the part in which they choose not to mask the top-1 probability:
```
sorted_indices_to_remove[..., 1:] = sorted_indices_to_remove[..., :-1].clone()
```

szha: Sorry for the confusion. I see that both sort and argsort are implemented but I don't see a way to get both values and indices in one call. The usage of `topk(k=-1)` that assumes the return values to be sorted seems to be undocumented, which is a bit of a concern.

src/gluonnlp/sequence_sampler.py
```diff
@@ -717,11 +717,20 @@ def forward(self, samples, valid_length, outputs, scores, step, beam_alive_mask,
...
+                 sorted_probs,
+                 mx.np.ones_like(probs)
+             )
+             # choose the borderline prob
+             p_prob = mx.np.min(masked_probs, axis=2, keepdims=True)
+             probs = mx.np.where(
+                 probs >= p_prob,
+                 probs,
+                 mx.np.zeros_like(probs)
+             )
```
sxjscience, line 733: The major difference between the current implementation and the original pytorch-based implementation is that when ``sampling_topp < max(probs)``, it is not clear which probability will be picked.

The pytorch-based implementation will always choose the token that are most probable.

PR #1513, opened 2021-01-31
[GLUON] turn off deduplicate option in save_parameters

szha:
## Description ##
this is a proposal to turn off deduplicate option in save_parameters

### Essentials ###
- [x] Code is well-documented

### Changes ###
- [x] this is a proposal to turn off deduplicate option in save_parameters

## Comments ##
- Turning on deduplication is known to cause issue when a model with shared parameters is saved and then loaded in a model where the parameter that is removed gets used.

@sxjscience @leezu feel free to comment if there's any concern with such change.

cc @dmlc/gluon-nlp-team

### DIFF
--- a/scripts/conversion_toolkits/convert_electra.py
+++ b/scripts/conversion_toolkits/convert_electra.py
@@ -400,15 +400,15 @@ def convert_tf_model(model_dir, save_dir, test_conversion, model_size, gpu, elec
                     ele_valid_length = valid_length[i]
                     assert_allclose(contextual_embedding[i, :ele_valid_length, :].asnumpy(),
                                     tf_contextual_embedding[i, :ele_valid_length, :], 1E-3, 1E-3)
-            model.save_parameters(os.path.join(save_dir, 'model.params'), deduplicate=True)
+            model.save_parameters(os.path.join(save_dir, 'model.params'))
             logging.info('Convert the backbone model in {} to {}/{}'.format(model_dir,
                                                                             save_dir, 'model.params'))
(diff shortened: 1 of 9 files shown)


sxjscience:
I think the previous design is that `deduplicate=True` will be turned on by default in MXNet: https://github.com/apache/incubator-mxnet/issues/18116#issuecomment-616807687.

The problem here is that the size of the artifact will be increased without deduplicate. However, it's also possible to keep the old design and still provide the user with a way to shrink the size of the dumped model parameters.

sxjscience:
For this PR, we will need to reconvert the model parameters and update the model files in GluonNLP.

leezu:
Given the use-case of loading parameters to models with a different set of shared parameters (eg gluonnlp), I don't intend changing the default in MXNet 2. For GluonNLP, it's fine to opt-in parameter deduplication. GluonNLP just need to modify the parameter loading code when attempts to load parameters to a different model with a different set of shared parameters.

leezu (commented):
Why not change the loading code instead? GluonNLP tries to load the parameter to a different model. But unlike the generic code in MXNet, GluonNLP does know exactly how the models differ and it should be straightforward to change the names (keys) in the parameter dictionary accordingly.

szha:
@leezu changing the loading logic sounds reasonable. We should make it so that users don't have to manage this, so it needs to be part of the API (maybe get_backbone? or a common parent class for backbone networks?). We should also keep it easy to customize the model for downstream tasks.

I will close this for now

PR #1514, opened 2021-02-02
[SCRIPT] update default arguments and add seed for scripts/generation/calculate_metrics.py

szha:
## Description ##
remove num_samples arguments in generation calculate_metrics. the number of samples can be inferred from the sample file.

### Essentials ###
- [x] Code is well-documented

### Changes ###
- [x] remove num_samples/num_bleu_samples from scripts/generation/calculate_metrics.py

## Comments ##
- I removed num_bleu_samples together with num_samples, as it's unclear to me whether subsampling is needed when calculating self BLEU. Feel free to comment or suggest otherwise.

cc @dmlc/gluon-nlp-team

### DIFF
--- a/scripts/generation/calculate_metrics.py
+++ b/scripts/generation/calculate_metrics.py
@@ -16,9 +16,8 @@ from gluonnlp.models.gpt2 import get_pretrained_gpt2
 def parse_args():
     parser = argparse.ArgumentParser(
         description='Calculate metrics for the generated sentences')
-    parser.add_argument('--file', type=str, required=True, help='Model name')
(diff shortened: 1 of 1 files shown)


sxjscience:
I think it's added because a straight-forward implementation of self-bleu has O(N^2) complexity. It will be too slow if we calculate it among lots of samples.

szha:
@sxjscience thanks for the comment. I suspected that performance was the reason. Still, I think it might be confusing to randomly subsample reference sets which results in variance in the scores from different runs. Should we switch to deterministic samples instead?

sxjscience:
We may fix the seed then.

sxjscience (approved):

PR #1521, opened 2021-02-10
[0.x] Update doc

szha:
## Description ##
Update doc after mx 1.7.

### Essentials ###
- [x] Code is well-documented

### Changes ###
- [x] Update doc after mx 1.7.

cc @dmlc/gluon-nlp-team

### DIFF
--- a/.pytype.cfg
+++ b/.pytype.cfg
@@ -5,4 +5,4 @@ inputs =
     src/gluonnlp
 
 # Python version (major.minor) of the target code.
-python_version = 3.5
+python_version = 3.6
--- a/docs/examples/sentence_embedding/bert.md
+++ b/docs/examples/sentence_embedding/bert.md
@@ -336,17 +336,6 @@ def calibration(net, dev_data, num_calib_batches, quantized_dtype, calib_mode):
                                                             logger=None)
     print('Calibration done with success.')
     return quantized_net
-
-# will remove until mxnet 1.7 release.
-try:
-    quantized_net = calibration(bert_classifier,
-                                dev_dataloader,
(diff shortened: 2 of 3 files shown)


leezu:
Why 3.8? People may be using the code with 3.6 already and it may break them?

szha:
OK I will stay with 3.6. 3.5 seems too old.

leezu:
Note that MXNet 1.6 pip wheels is incompatible with Horovod https://github.com/horovod/horovod/issues/2425

ci/prepare_clean_env.sh
```diff
@@ -20,9 +20,10 @@ conda activate ./conda/${env_name}
 conda list
 printenv
 
-pip install cmake
-pip install -v -e .
-# pip install horovod --no-cache-dir -U
-python -m spacy download en
-python -m spacy download de
-python -m nltk.downloader all
+python3 -m pip install awscli
```
barry-jin, line 23: We can install awscli through conda-forge `conda install -c conda-forge awscli`. This may fix the issue in CI.

szha: I realized that the issue happens on the batch job and it was the ci/batch/docker/gluon_nlp_job.sh I should be updating.

barry-jin: We are using the CI docker in master branch here and the docker setting in v0.x branch are currently not used for v0.x CI. I keep the docker setting and scripts here for some future work that we create a job queue on batch for v0.x CI so that we can maintain the CI dockers seperately.

leezu (approved):

PR #1532, opened 2021-02-25
PyTorch attention cells, transformer, bert and conversion script

leezu:
Add PyTorch attention cells, transformer, bert and conversion script. Correctness verified via included tests and conversion script.


sxjscience (approved):
I'll approve. We can improve the testing later since this is the initial version of the pytorch-based GluonNLP.

PR #1581, opened 2021-10-06
Quantize QuestionAnswering models

bgawrych:
## Description ##
This PR enables quantization on question answering scripts.
Added custom calibration collector to avoid significant accuracy drop

### DIFF
--- a/scripts/question_answering/run_squad.py
+++ b/scripts/question_answering/run_squad.py
@@ -120,7 +120,7 @@ def parse_args():
                              'this will be truncated to this length. default is 64')
     parser.add_argument('--pre_shuffle_seed', type=int, default=100,
                         help='Random seed for pre split shuffle')
-    parser.add_argument('--round_to', type=int, default=None,
+    parser.add_argument('--round_to', type=int, default=8,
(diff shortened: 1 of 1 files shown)


bartekkuncer (commented):
Looks good to me.

scripts/question_answering/models.py
```diff
@@ -287,10 +289,14 @@ def forward(self, tokens, token_types, valid_length, p_mask, start_position):
             Shape (batch_size, sequence_length)
         answerable_logits
         """
+        backbone_net = self.backbone
+        if self.quantized:
```
anko-intel, line 293:
```suggestion
        if self.quantized_bacbone is not None:
```
end remove quantized ?

anko-intel (approved):

scripts/question_answering/models.py
bgawrych: I thought about it, but I remained this quantized flag as switch on/off of quantized model - not sure if it is really usable. What do you think?

bartekkuncer (approved):

PR #1587, opened 2022-04-07
Add assert for doc_stride, max_seq_length and max_query_length

bartekkuncer:
## Description ##
This change adds assert for doc_stride, max_seq_length and max_query_length as incautious setting of them can cause data loss when chunking input features.

cc @dmlc/gluon-nlp-team

### DIFF
--- a/scripts/question_answering/run_squad.py
+++ b/scripts/question_answering/run_squad.py
@@ -41,6 +41,7 @@ mx.npx.set_np()
 CACHE_PATH = os.path.realpath(os.path.join(os.path.realpath(__file__), '..', 'cached'))
 if not os.path.exists(CACHE_PATH):
     os.makedirs(CACHE_PATH, exist_ok=True)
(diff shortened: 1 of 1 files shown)


scripts/question_answering/run_squad.py
```diff
@@ -151,6 +152,7 @@ def parse_args():
                              'use --dtype float16, amp will be turned on in the training phase and '
                              'fp16 will be used in evaluation.')
     args = parser.parse_args()
+    assert args.doc_stride <= args.max_seq_length - args.max_query_length - SEPARATORS, 'Possible loss of data while chunking input features'
```
bgawrych, line 155:
```suggestion

    assert args.doc_stride <= args.max_seq_length - args.max_query_length - SEPARATORS, \
                             'Possible loss of data while chunking input features'

```

bgawrych (approved):

